PR #1, opened 2017-04-06
Render the apb LABEL in the dockerfile instead of using a template

rthallisey:
Instead of using a template to render the APB Dockerfile, we should
substitute the encoded apb config into that field. This will make it
so we don't have to anticipate how the developer will build their
Dockerfiles.


fabianvf:
Would it be possible to make it a flag instead? It seems useful to me to be able to generate a working dockerfile with the tooling, particularly for someone who is just getting started with asb/abp stuff.

rthallisey:
I can make a flag, but I think we should move away from generating a working dockerfile because we'll have to keep in sync the dockerfile template and the dockerfiles from the library.  This issue gets even trickier if as a dev I start customizing my images to look different than that template.  My PR https://github.com/fusor/ansible-playbook-bundle-library/pull/1 highlights this because ```ansibleapp prepare``` constantly overwrites the Dockerfile changes I made in that patch. 

What if I created a new command ```ansibleapp init``` that created the working dockerfile?

fabianvf:
Yeah, init makes a lot more sense, you're right that this behavior doesn't really belong in the prepare command.

jmrodri:
@rthallisey @dymurray @eriknelson can you guys review this and decide if this is worth fixing and moving forward?

rthallisey:
If we're still using the engine to encode our yaml data, then I think it's still relevant.  I'll rebase.

eriknelson:
+1, it's still relevant.

The intention for the template was to steer people in the right direction because we're opinionated about how your image needs to look to be a valid apb. There's still value in generating a basic starting point with some kind of `init` command if there's nothing there, but under no circumstance should we be blowing away people's Dockerfiles, which I was guilty of :sweat_smile:.

We also still need a tool to assist in generating that blob from the apb.yaml and injecting it into the correct label without disrupting developer customization.

I'd be fine throwing out what we've got if there's a better way to organize the workflow, but I do think having the above two features are important (the second being a necessity).

dymurray:
@rthallisey I have some changes in my branch here: https://github.com/dymurray/ansible-playbook-bundle/tree/tooling which I have been working in to get 'init' and 'prepare' working as well as covering the needs on the developer experience cards we have in progress. I am happy to look at what you have here because I wasn't sure what the best way to get this portion done but I believe what I have was working pretty effectively in my tests. 

I'm currently based off master so would be happy to figure out the best way to get our work integrated. We can find a time to discuss this.

rthallisey:
@dymurray ok cool, I'll take a look.

jmrodri:
@rthallisey @fabianvf @dymurray is this PR still relevant? If so, let's rebase.

rthallisey:
Closing in favor of https://github.com/fusor/ansible-playbook-bundle/pull/18 and https://github.com/fusor/ansible-playbook-bundle/issues/13

PR #5, opened 2017-04-25
merge in some of the changes from apb branch

jmrodri:
There were some minor things that I did in the apb branch that were not captured with the resent renaming work. I've included them here.

Biggest consequence was using PYTHON 2 in the rpm, and adding the man pages to it.


johnkim76:
ACK!

PR #8, opened 2017-04-27
fix blob chunking when mod == 0

jmontleon:
Needs more testing, but after a couple tests it seems ok.


eriknelson:
ACK, thank you @jmontleon, clearly needed more coffee :)

PR #19, opened 2017-05-15
Adding specific example for empty paramaters.

shawn-hurley:


rthallisey:
+1

docs/design.md
```diff
@@ -30,6 +30,15 @@ parameters:
     type: int
     default: 3306
 ```
+Or with an empty parameters field
+```yaml
+name: fusor/etherpad-apb
+description: Note taking web application
+bindable: true
+async: optional
+parameters: []
```
rthallisey: Not everything has a default for example ```db_name```.  Will it still work without setting it?

rthallisey: Actually, the defaults for these values are only known by this file.  What ends up happening if you don't pass any params?

shawn-hurley: I should change this one to one that has no parameters. Do we have an APB that would work for no parameters? Right now RDS - MySql is set up that way but is there a better example that I should choose?

rthallisey: I'm not sure if there's one that works with no params.  But, we definitely want to put an example here that does because folks will look at it and try to use it.

shawn-hurley: We could create a placeholder example, call it 
```yaml
name: example/no-parameters 
description: Example showing no parameters
bindable: false
async: optional
parameters: []
```

And keep an issue open to update once we create an apb that does take no parameters?

rthallisey: Sure, a place holder works.  I agree with keeping an open issue until we have an example apb to replace this place holder.

rthallisey:
+1

PR #22, opened 2017-05-18
Template out actions for playbooks and roles

dymurray:


src/README.md
```diff
@@ -28,14 +28,14 @@ an `apb.yaml`, a `Dockerfile`, a `roles` directory, and a `playbooks` directory.
 APB name.
 
 #### Options:
-
+```
 `--organization ORG` specifies an organization name for the APB image on the `Dockerfile`. Defaults to `ansibleplaybookbundle`.
 `--async ASYNC_TOPION` specifies the asynchronous operation of the APB. Defaults to `optional`
```
jmrodri: s/TOPION/OPTION

I still think `--async` should be a boolean instead of having the user give a specific value.

src/apb/cli.py
```diff
@@ -42,14 +42,20 @@ def subcmd_init_parser(parser, subcmd):
     )
 
     subcmd.add_argument(
-        '--bindable', action='store_true', dest='bindable',
-        help=u'Make application bindable', default=False
+        '--not-bindable', action='store_false', dest='bindable',
+        help=u'Make application not bindable on the spec.', default=True
```
jmrodri, line 49: I think I missed something. Does this mean `--bindable` is no longer valid? or does this mean *both* `--bindable` and `--not-bindable` are valid?

src/apb/engine.py
```diff
@@ -117,6 +120,43 @@ def insert_encoded_spec(dockerfile, encoded_spec_lines):
...
+    write_file(playbook_out, action_path, True)
+
+
+def generate_playbook_files(playbooks_path, roles_path, bindable, skip, apb_name):
+    ENV = Environment(loader=FileSystemLoader(DAT_PATH))
+    template = ENV.get_template(EX_PLAYBOOK_FILE)
+
+    print("Generating playbook files")
+
+    if 'provision' not in skip:
```
rthallisey: Could do a for loop here that iterates through a list of actions [provision, deprovision, bind, unbind]  since the only difference between these if statements is the $ACTION.

src/apb/engine.py
```diff
@@ -208,9 +248,25 @@ def cmdrun_init(**kwargs):
...
 
     description = "This is a sample application generated by apb init"
 
+    # Ensure that command line options are acceptable
+    for skip_option in skip:
+        if skip_option not in SKIP_OPTIONS:
+            raise Exception(
+                "Error: [%s] is not a valid option to skip. The following are skippable options: %s" %
+                (skip_option, SKIP_OPTIONS)
+            )
```
jmrodri: I would've done `--no-provision` `--no-deprovision` etc vs `--skip provision` or `--skip-provision`, `--skip-deprovision`. This avoids having to do the validation of user input. You could still do it in an automated fashion so that you can add skippable items without having to duplicate a bunch of code.

```
SKIP_OPTIONS = ['provision', 'deprovision', 'bind', 'unbind', 'roles']

for opt in SKIP_OPTIONS:
    subcmd.add_argument(
        '--skip-%s' % opt, action='store_false', dest='skip-%s' % opt,
        help=u'Specify which playbooks to not generate by default.'
    )
```
I *think* that will work. But then the user doesn't have to remember to type in the correct string and you can remove the validation altogether. It also allows you to add more to SKIP_OPTIONS in the future without having to update the generation code.

src/apb/engine.py
```diff
@@ -208,9 +248,25 @@ def cmdrun_init(**kwargs):
...
+
+    if async not in ASYNC_OPTIONS:
+        raise Exception(
+            "Error: [%s] is not a valid option for async. The following are asynchronous options: %s" %
+            (async, ASYNC_OPTIONS)
+        )
+
+
```
jmrodri: I'd probably do the same thing with ASYNC as I suggested with SKIP above.

src/apb/engine.py
jmrodri: The idea I'm shooting for is instead of:
```
--skip provision
```
They can type in
```
--skip-provision
```
While it seems silly for a single dash but it makes it less error prone from the users perspective.

jmrodri (changes requested):
If my suggestions will cause too much pain, we can chat. Or we can create a card if it's something we want to do but not now. I prefer to limit the amount of user input we have to deal with especially if we have fixed values we are looking for.

src/apb/engine.py
```diff
@@ -208,9 +248,25 @@ def cmdrun_init(**kwargs):
     bindable = kwargs['bindable']
     async = kwargs['async']
     params = kwargs['params']
+    skip = kwargs['skip'] or []
 
     description = "This is a sample application generated by apb init"
 
+    # Ensure that command line options are acceptable
+    for skip_option in skip:
```
rthallisey: Instead of validating the skip params here, you could use choices in argparse to limit what the user can enter.
https://docs.python.org/3/library/argparse.html#choices

rthallisey (changes requested):

src/apb/engine.py
dymurray: @rthallisey THANK YOU! I knew there was a way to do this with argparse and it was driving me nuts I couldn't remember.

src/apb/cli.py
dymurray: @jmrodri we wanted to make the application bindable by default, with a flag to disable binding. --bindable is no longer valid but the 'bindable' variable is defaulted to True.

jmrodri (approved):

src/apb/engine.py
```diff
@@ -129,31 +130,24 @@ def generate_playbook_files(playbooks_path, roles_path, bindable, skip, apb_name
...
-        if 'roles' not in skip:
-            os.mkdir(os.path.join(roles_path, 'bind-' + apb_name))
+    print("Generating playbook files")
 
-    if 'unbind' not in skip:
-        # Create Unbind playbook
-        write_playbook_file(playbooks_path, UNBIND_PLAYBOOK, template, apb_name, 'unbind')
-        if 'roles' not in skip:
-            os.mkdir(os.path.join(roles_path, 'unbind-' + apb_name))
+    for playbook in playbooks:
```
rthallisey: You can remove the ```playbook``` list and do ```for playbook in playbooks_dict.keys():```.  That will loop through the dict keys and will get the same functionality as using the list.

rthallisey (changes requested):

rthallisey (approved):

rthallisey:
+1

PR #24, opened 2017-05-19
Updated README and reference to old code

dymurray:


jmrodri (changes requested):
The changes look ok. But the original files has pip mentions at the top. Remove those since they are already documented in the root README file.

dymurray:
@jmrodri Updated the README. Also made a change to restamp the spec when building.

jmrodri (approved):

PR #27, opened 2017-05-22
Changed extensions from yaml to yml

dymurray:


jmrodri (approved):
:+1: good to be consistent.

PR #30, opened 2017-05-22
Add GPL V2 license for APB repo

dymurray:


cfchase:
ACK

rthallisey:
@dymurray  Would this mean that any APB written carries the GPL licence?  If so, then projects won't want to keep them in their repos because I think it contaminates that project's licence?

dymurray:
@rthallisey From my understanding it would not mean that any APB written using this program carries the GPL license by default. @jmrodri can you comment?

rthallisey:
https://opensource.stackexchange.com/questions/2501/does-using-a-gplv2-code-generator-force-my-project-to-be-similarly-licensed

>According to the GPL FAQ the output of a GPL program is not licensed under the GNU GPL, unless it copies substantial parts of source code into the output which are complex enough to fall under copyright:

    Q:In what cases is the output of a GPL program covered by the GPL too?

    A: Only when the program copies part of itself into the output.
>

rthallisey:
I don't think we're copying substantial parts of source code.  So I think it should be ok.

dymurray:
@rthallisey Yeah agreed, thanks for the source.

PR #31, opened 2017-05-22
update for @rthallisey suggestions from PR#25

johnkim76:


docs/developers.md
```diff
@@ -10,11 +10,12 @@ In order to create an APB, you will need to start with a skeleton APB directory
...
 
-Running the `apb init` will create a APB directory structure shown below:
+Run the `apb init` command
 ```bash
 $ apb init my-apb
-
-$ tree my-apb
+```
+The following APB directory structure will be created:
+```bash
```
rthallisey: I saw in the design doc we have a sub field that has ```##### Directory Structure```.   Maybe we can do something like that?
https://github.com/fusor/ansible-playbook-bundle/blob/master/docs/design.md#directory-structure

rthallisey (changes requested):

docs/developers.md
johnkim76: when the `apb init my-apb` is ran, the folder structure is as shown.  It's also documented in the [getting started ](https://github.com/fusor/ansible-playbook-bundle/blob/master/docs/getting_started.md#creating-your-first-apb)

johnkim76: if that section is not clear, should I just mention to run the `apb init` coommand and remove showing the directory structure?

rthallisey: I think the section is clear. I like the addition of the directory structure and I think it belongs. My suggestion is just the addition of the title to follow the same pattern from the design doc. IMO with or without it works for me.

johnkim76: Oh, I see, yes, I will add the title

rthallisey (approved):

docs/developers.md
```diff
@@ -93,7 +96,6 @@ And a simple `deprovision.yml` may look like this.
 We will also need to create the Ansible roles as specified in the actions. The below directory structure shows what it can look like:
 
 ```bash
-$ tree my-apb
 my-apb/
```
rthallisey, line 100: same as above:. ```### Directory Structure```

johnkim76: thanks, I've made the updates

rthallisey:
👍

PR #32, opened 2017-05-22
Update setup.py and add setup.cfg for publishing to pypi

dymurray:


jmrodri (approved):
Looks good. Can we tag it with a tito formatted tag?
`git tag apb-0.1.0`

We might be able to do simultaneous releases to fedora, centos, rhel, & pypi all using `tito release` in the future.

jmrodri:
We can also rename the specfile to be ansible-playbook-bundle.spec so the package will be called ansible-playbook-bundle vs apb. But we can deal with that later.

PR #34, opened 2017-05-23
Jinja2 is required for `apb init` to work properly

flaper87:


jmrodri:
We need to make sure we're not getting lucky because we have a dep installed locally.

dymurray (approved):

PR #40, opened 2017-05-30
Generate roles/tasks/main.yml files on apb init.

cfchase:


src/apb/engine.py
```diff
@@ -19,15 +21,43 @@
...
+EX_DOCKERFILE = 'Dockerfile.j2'
 EX_DOCKERFILE_PATH = os.path.join(DAT_PATH, EX_DOCKERFILE)
 
-PLAYBOOKS_DIR = 'playbooks'
-EX_PLAYBOOK_FILE = 'playbook.yml.j2'
-PROVISION_PLAYBOOK = 'provision.yml'
-DEPROVISION_PLAYBOOK = 'deprovision.yml'
-BIND_PLAYBOOK = 'bind.yml'
-UNBIND_PLAYBOOK = 'unbind.yml'
+ACTION_TEMPLATE_DICT = {
```
dymurray, line 27: +1 Much cleaner

src/apb/dat/Dockerfile.j2
```diff
@@ -1,5 +1,4 @@
 FROM ansibleplaybookbundle/apb-base
-# MAINTAINER {{ $MAINTAINER }}
 
 LABEL "com.redhat.apb.version"="0.1.0"
```
dymurray, line 3: This doesn't need to be addressed in this PR but I'm wondering if the version should be generated by the tooling for the Dockerfile. Technically we are up to 0.1.1 in PyPi and releases on Github I'm not sure if this version number should correlate to the tooling version number or just referring to the spec version of the APBs being published.

I'll discuss this with @cfchase and we'll address it later.

dymurray (approved):

dymurray:
Tested this myself and everything looks good. Great work on cleaning up some of my code :+1:

PR #45, opened 2017-06-06
Provide better source installation instructions

jmontleon:
We should not be suggesting or encouraging the use of `sudo pip install`. It will not be safe to use until at least Fedora 27.

https://fedoraproject.org/wiki/Changes/Making_sudo_pip_safe


cfchase (approved):
ACK

PR #46, opened 2017-06-06
add python-jinja2 to rpm spec and drop ver for EL7 and F25

jmontleon:
I was just installing the rpm on a clean Fedora 25 system and received an error because jinja2 was missing.


dymurray:
ACK, thank you Jason.

PR #54, opened 2017-06-16
[DO NOT MERGE] add python3 deps and fix python requests dep

jmontleon:
@dymurray are we able to drop python-requests to 2.6.0 to satisfy the dependency using the package already available in RHEL/CentOS 7? If so I'll amend this PR to drop the version in the requirements file and the rpm spec.


dymurray:
@jmontleon yes that python-requests package is fine. That looks good to me. ACK.

PR #56, opened 2017-06-16
APB init modifications

cfchase:
* Made --org optional if you include the org in your apb init name such as  `apb init cfchase/my-apb`
* Made bindable default to false since users will be creating consumer apps way more often than bindable services.


src/apb/engine.py
```diff
@@ -283,19 +283,27 @@ def update_dockerfile(spec_path, dockerfile_path):
...
+        'bind': kwargs['skip-bind'] or not kwargs['bindable'],
+        'unbind': kwargs['skip-unbind'] or not kwargs['bindable'],
         'roles': kwargs['skip-roles']
     }
 
+    apb_name_arr = kwargs['name'].split('/')
+    apb_name = apb_name_arr[-1]
+
+    organization = kwargs['org']
+    if organization is None:
```
dymurray, line 301: +1 This makes it easier.

dymurray:
ACK

PR #57, opened 2017-06-21
APB tooling improvements init/push

cfchase:
Added ID generation on init and build.
Allow apb push without specifiying broker if user is logged in using oc.


dymurray (approved):
Just tested using the new init and push function with success. 'apb build' also encapsulates prepare now. Everything works as expected. ACK.

PR #59, opened 2017-06-23
Update deprovision template about removing a deployment config, not its resources

codificat:
When removing a DeploymentConfig, OpenShift will take care to remove its associated resources like Replication Controllers and deployed pods, there's no need to manually clean these up.

### DIFF
--- a/src/apb/dat/roles/deprovision/tasks/main.yml.j2
+++ b/src/apb/dat/roles/deprovision/tasks/main.yml.j2
@@ -26,25 +26,11 @@
 
 ##############################################################################
 ## Deprovision a deployment config
-## Unlike other resources, deprovisioning deployment configs also requires
-## cleanup of additional resources created.  This includes the pods and
-## replication controller
+## When removing a Deployment Config, OpenShift will automatically clean up
+## its associated resources like replication controllers and pods
 ##############################################################################
-## Have to scale down first or else orphaned pods will be left.
-#- openshift_v1_deployment_config:
-#    name: {{ apb_name }}
-#    namespace: '{% raw %}{{ namespace }}{% endraw %}'
-#    replicas: 0
-#    state: present
-
 ## Delete the deployment config.
 #- openshift_v1_deployment_config:
 #    name: {{ apb_name }}
-#    namespace: '{% raw %}{{ namespace }}{% endraw %}'
-#    state: absent
-
-## Destroy the orphaned replication controller.
-#- k8s_v1_replication_controller:
-#    name: {{ apb_name }}-1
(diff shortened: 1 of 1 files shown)


jwmatthews:
@dymurray, @cfchase, or @fabianvf  Please review when you can, I thought you guys hit some issues with deprovision of deployment config.

fabianvf:
Openshift actually doesn't handle cleaning up those resources. We're using the REST API, so we don't have the client side logic in oc that handles deleting the orphaned resources. Unless that's changed, did you verify that the resources were cleaned up properly?

codificat:
AFAIK it is supposed to work like this... a [quick look at the code](https://github.com/openshift/origin/blob/v3.6.0-alpha.2/pkg/deploy/cmd/delete.go#L42-L45) also seems to suggests this. I might be missing something though...

I only did a basic sanity check: deleted a DC and observed how the corresponding RCs and pods went away too. Are you not seeing this then?

jwmatthews:
@codificat did you try the delete of DeploymentConfig with the ansible k8s modules or with an oc cmd.

I think the link you gave in prior comment is for the oc command.

@fabianvf was mentioning how the oc command has client side logic to handle the cleanup, but the actual REST APIs in k8s do not.  

The ansible modules communicate at REST API layer, bypassing oc client, hence behavior is a little different.

APBs can be written to use either oc commands are valid, as are the newer ansible k8s modules.

Perhaps we should document this distinction?

codificat:
@jwmatthews @fabianvf right, I see your point: the `cascade` option in `oc` (default to true) in action.

Still, I tested a removal of a DC via the ansible modules (`openshift_v1_deployment_config.state: absent` and no other associated logic) and this did also remove associated resources...

Also, an `oc delete --cascade=false dc/mydc` will only call the DELETE API call for the DC (no additional logic), but after that the DC is deleted its related RCs/pods get reaped too...

jwmatthews:
@codificat sounds promising like we can simplify things, thanks for bringing this up.

@dymurray @fabianvf would one of you please test this out, see if behavior has changed since we last attempted?

fabianvf:
@jwmatthews that would be exciting, @dymurray could you test it? I'm still on the installer stuff right now.

codificat:
Some more details: automatic clean up of a DC's related resources is a new feature in 3.6, available as of [origin 3.6alpha2](https://github.com/openshift/origin/releases/tag/v3.6.0-alpha.2) which is what I was testing with.

Maybe it's a bit too early to change the deprovision template then?

cfchase:
Thanks @codificat .  Tested and worked for me.

ACK

dymurray:
Thank you @codificat it worked for me as well. I am good to merge this.

PR #63, opened 2017-07-03
Fix the help description for list subcommand

ghost:


jmrodri (approved):
Thanks for the patch.

jmrodri (approved):

PR #67, opened 2017-07-14
Add dependency tooling

dymurray:


dymurray:
[(5) APB Image Dependencies - Phase I](https://trello.com/c/pamKcgbw/305-5-apb-image-dependencies-phase-i)

dymurray:
[(13) APB Image Dependencies: STRETCH GOAL: Develop a P.O.C. for inspecting an APB and determining/updating the referred images automatically.](https://trello.com/c/hzkmboR5/306-13-apb-image-dependencies-stretch-goal-develop-a-poc-for-inspecting-an-apb-and-determining-updating-the-referred-images-automati)

docs/design.md
```diff
@@ -55,25 +55,51 @@ description: "My New APB"
...
+      title: Root Password
+      description: root password for mariadb 
+      type: string
+      default: admin
+required:
+  - mariadb_name
+  - mariadb_user
+dependencies:
+  - docker.io/mariadb:latest
+  - docker.io/tvelocity/etherpad-lite:latest
```
rthallisey, line 99: Few questions:
Does order matter here?
Why would the etherpad-apb depend on etherpad-lite? 
Shouldn't we depend on the mariadb-apb instead of mariadb?

src/apb/engine.py
```diff
@@ -291,14 +292,29 @@ def touch(fname, force):
...
     spec_path = os.path.join(project, SPEC_FILE)
+    roles_path = os.path.join(project, ROLES_DIR)
 
     # ID specfile if it hasn't already been done
     if 'id' not in spec:
         gen_spec_id(spec, spec_path)
 
+    if not ignore_deps:
+        expected_deps = load_source_dependencies(roles_path)
+        if 'required' not in spec:
```
rthallisey, line 306: Why is a 'required' and 'metadata' check inside the ignore_deps if block?

src/apb/engine.py
```diff
@@ -321,6 +337,14 @@ def update_dockerfile(project):
     print('Finished writing dockerfile.')
 
 
+def load_source_dependencies(roles_path):
+    print('Trying to guess list of dependencies for APB')
+    output = subprocess.check_output("/bin/grep -R \ image: "+roles_path+"|awk '{print $3}'", stderr=subprocess.STDOUT, shell=True)
```
rthallisey, line 342: Are we making any suggestions for dependencies? How are we choosing what needs what deps?

src/apb/engine.py
```diff
@@ -440,13 +466,33 @@ def cmdrun_init(**kwargs):
...
+
+    specfile_out = load_example_specfile(apb_dict, spec['parameters'])
+    write_file(specfile_out, spec_path, True)
     update_dockerfile(project)
 
 
 def cmdrun_build(**kwargs):
     project = kwargs['base_path']
-    spec = update_spec(project)
+    ignore_deps = kwards['ignore_deps']
```
rthallisey: s/kwargs/kwards/

rthallisey (changes requested):

docs/design.md
dymurray: These images are the images referenced in the provision role. Our etherpad-apb launches two containers using those images so they are listed as dependencies in no specific order.

src/apb/engine.py
dymurray: This block of code is only executed if the ignore_deps flag isn't set. There's no need to go through the templating process if the user skips it. These checks are so the templating process doesn't break.

src/apb/engine.py
dymurray: By choosing which images are referenced in the k8s ansible modules or the openshift modules. If the APB specifies container images then we want to be able to intuitively pick them out.

rthallisey: I had a different definition of dependencies.  These are the images that an apb consumes. That's where I got confused.

dymurray:
@rthallisey Fixed the typo I had in my commit.

rthallisey (approved):

PR #71, opened 2017-07-19
Addding insecure option for bootstrap,push,list,remove

shawn-hurley:
* Necessary due to merge of https for the broker route.

@dymurray The loading of the certificate and attempting to use that, will be in a subsequent PR. 
I think that this is block @johnkim76 so I want to get him something ASAP. 

Also, I think that the open shift is already defaulting to insecure, so I could see us amending this to do that as well?

### DIFF
--- a/src/apb/cli.py
+++ b/src/apb/cli.py
@@ -27,6 +27,12 @@ def subcmd_list_parser(subcmd):
         dest='broker',
         help=u'Route to the Ansible Service Broker'
     )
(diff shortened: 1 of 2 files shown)


johnkim76:
@shawn-hurley I'll test this out thank you.

johnkim76:
ACK 
@shawn-hurley i've successfully tested the '-k' insecure option.

johnkim76 (approved):

PR #83, opened 2017-07-31
simplify perms

tchughesiv:
we don't need a `chown` here, `chmod -R g=u` is sufficient since `apb` user is in root group.


tchughesiv:
corresponding changes made to examples in this PR:
https://github.com/fusor/apb-examples/pull/99

rthallisey (approved):

PR #96, opened 2017-08-17
Added validation checks on the spec file

dymurray:


jdob (commented):
For what it's worth, you can save a ton of boiler plate code in a number of ways. For instance, you could have a list of required keys and loop over those, similar to:

```
REQUIRED_KEYS = ['name', 'description', 'bindable']
for k in REQUIRED_KEYS:
  if k not in spec:
    print("Spec is not valid. %s field not found" % k)
    return False
```

Something else you may want to consider is not returning on the first failed test and instead outputting all of the issues at once. That'll save frustration on the user's part for having to individually step through each issue.

dymurray:
Thanks @jdob I updated it with your advice. Much cleaner :)

cfchase (approved):
ACK

jmrodri (approved):
looks good. This would've been a prime candidate for a unit test.

PR #101, opened 2017-08-23
Added unit testing setup and skeleton

jdob:
Nosetests will be installed as a testing requirement. Tests can be run by simply running "nosetests" in the root of the project.

I didn't see a section on actively developing on the abp project, so I added a quick note about the test requirements file in the "Installing from source" section.

I don't typically see a "src" directory in Python applications so I was a bit unsure of where to add the tests directory, but I threw it in there figuring that made the most sense.

I added a simple test to show the testing setup is valid, but that test will need to be flushed out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,13 @@ Install requirements and run the setup script (requires python)
 ```
 pip install -r src/requirements.txt && python setup.py install
 ```
+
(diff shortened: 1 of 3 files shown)


dymurray:
ACK

PR #104, opened 2017-08-24
Add APB testing to the apb tool

shawn-hurley:
Allows for users to more easily test the APB defined the [proposals](https://github.com/fusor/ansible-playbook-bundle/blob/master/docs/proposals/testing_implementation.md).

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -20,6 +20,7 @@ from ruamel.yaml import YAML
 from openshift import client as openshift_client, config as openshift_config
 from jinja2 import Environment, FileSystemLoader
 from kubernetes import client as kubernetes_client
+from kubernetes.client.rest import ApiException
 
 ROLES_DIR = 'roles'
 
@@ -363,6 +364,7 @@ def get_asb_route():
         asb_route = None
     return asb_route
 
+
 def delete_controller_manager_pod():
     pod_name = None
     try:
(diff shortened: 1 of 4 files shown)


shawn-hurley:
Working on updating the getting_started docs.

src/apb/engine.py
```diff
@@ -632,3 +747,39 @@ def cmdrun_bootstrap(**kwargs):
...
+
+    print("Successfully built APB image: %s" % tag)
+
+    # run image that was just created.
+    create_image_pod(tag)
+    test_result = retrieve_test_result()
+    test_results = []
+    if test_result is None:
+        print("Unable to retrieve test result.")
+        return
```
dymurray, line 775: Should this exit instead of return? I'm assuming this means it failed the test?

src/apb/engine.py
```diff
@@ -379,6 +381,119 @@ def delete_controller_manager_pod():
...
+                'namespace': 'default',
+            }],
+            'roleRef': {
+                'name': 'cluster-admin',
+            },
+        }
+        api.create_namespaced_role_binding("default", role_binding)
+    except Exception as e:
+        api = openshift_client.OapiApi()
+        # HACK: this is printing an error but is still actaully creating the
```
dymurray: nit: actually*

src/apb/engine.py
```diff
@@ -632,3 +747,39 @@ def cmdrun_bootstrap(**kwargs):
...
+        tag = kwargs['tag']
+    print("Building APB using tag: [%s]" % tag)
+
+    client = docker.DockerClient(base_url='unix://var/run/docker.sock', version='auto')
+    client.images.build(path=project, tag=tag)
+
+    print("Successfully built APB image: %s" % tag)
+
+    # run image that was just created.
+    create_image_pod(tag)
```
dymurray, line 770: I would add a check here if the create_image_pod resulted in an exception and exit

src/apb/engine.py
shawn-hurley: As long as we call `clean_up_image_run` then yeah that makes sense to me. 

I want to make sure that we delete the things that we created so that they can re-run the `test` command without dealing with duplicate things, especially the pod. 
:+1: Will make that change

src/apb/engine.py
shawn-hurley: I am catching all the expectations for create_image_pod, and I think it will still need to run some clean up potentially. 
```python
def create_image_pod(image_name):
      try:                                                                         
          .....                                                
      except Exception as e:                                                       
          pod_name = None                                                          
          print("failed - %s" % e)
```

Would you prefer to not catch the exception here and instead catch it in the `cmdrun_test` function?

dymurray: Ah I see how it runs through the cleanup process if it fails so what you have makes sense. LGTM.

dymurray:
ACK

PR #117, opened 2017-09-22
[Proposal] Versioning of APBs

dymurray:


jmrodri:
AMEN! Excellent!

jmrodri:
> Does version match broker/OCP release or is it it's own?

I've always thought they should be their own unless we never want to release a broker without an OCP version bump. Then I could see the argument for them being the same.

docs/proposals/versioning.md
```diff
@@ -0,0 +1,25 @@
+## Plan for versioning APBs
+
+### Current status:
+* Currently all APBs are 0.1.0
+  * This is stored in a label on the image titled “com.redhat.apb.version”
+* Plan is to bump all APBs to a newer version to coincide with 3.7 Broker
+
+### Questions:
+* Should this be a major version bump? I.e. 1.0.0, or minor would be 0.2.0
```
jmrodri, line 9: +1 to doing a major version jump right now to something like 1.0.0.  A
minor version jump is not appropriate because minor versions should be
compatible with previous minor versions. That is they should not be a
breaking change.

docs/proposals/versioning.md
```diff
@@ -0,0 +1,25 @@
...
+* Do we tag the images with the version of the broker or APB spec version? Should they match?
```
jmrodri, line 10: Not sure if we do the versions by tags or if it's in the apb image name. The version will be in the label right? Can we change labels as easily as tags?

jmrodri:
General comments not specific to any line in the proposal:

So major (x stream) versions will mean we most likely will break
compatibility. Minor ( y stream) I'd assume should still work with
older brokers the way they used to but not necessarily gain the new
functionality. And the z stream bump is purely a non-breaking change to
the spec.

We need to define what's considered breaking changes (at least
conceptually) so we know what would require a major version bump. I
think plans was one example.

Same thing with minor and z-stream.

jmrodri:
I know John mentioned using semantic versioning, which is what I was
trying to describe above. http://semver.org/

Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible
manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as
extensions to the MAJOR.MINOR.PATCH format.

docs/proposals/versioning.md
dymurray: The version will be in the label yes, but the broker doesn't have an easy way to see that version without introspecting the manifest. I think it makes sense to tag the image with whatever version is on the label.

docs/proposals/versioning.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Thoughts:
+* Makes sense to do a major version bump because the version has not changed since we were ansibleapp.
+* Since the ‘plan’ format of the APB is not likely to change anytime soon it would make sense to establish the schema of the spec in a major bump. There are still some new APB developers who find old examples and try to use it without plans.
+* 1.x.x APBs work on ASB/OC 3.7
+* 0.x.x APBs are < ASB 3.6 
+* Image tags should match whatever version number we choose. This would be a pro for versioning APBs in the same vein as ASB. i.e. 3.7 ASB can launch images tagged with 3.7.
```
jmrodri, line 17: So there is definitely value here with keeping versions locked in with OCP it makes it very easy to see what works with what. 

But this also means that we can NOT make any major changes in a given release. Let's say at the beginning of 3.8 development we say that APB will now be versioned 3.8. If we do a change the equivalent to plans mid development cycle that would require us to go to 4.0 which means APB 4.0 would work with ASB/OCP 3.8 which is quite confusing.

If the versions were NOT locked down to OCP, then we could independently bump ASB/APB versions to ensure that we know APB x.y is compatible with ASB x.

dymurray: +1. I think you make good points for not locking the version to OCP. I'm in agreement as I expect the APB spec to develop major feature changes before OCP does.

docs/proposals/versioning.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Questions:
+* Should this be a major version bump? I.e. 1.0.0, or minor would be 0.2.0
+* Do we tag the images with the version of the broker or APB spec version? Should they match?
+
+### Thoughts:
+* Makes sense to do a major version bump because the version has not changed since we were ansibleapp.
+* Since the ‘plan’ format of the APB is not likely to change anytime soon it would make sense to establish the schema of the spec in a major bump. There are still some new APB developers who find old examples and try to use it without plans.
+* 1.x.x APBs work on ASB/OC 3.7
+* 0.x.x APBs are < ASB 3.6
```
djzager, line 16: nit: 0.x.x APBs are `<=` ASB 3.6 ?

docs/proposals/versioning.md
djzager: If I understand how the registry adapters are working already, then they are already using introspection to create the APB Spec struct. This is not me advocating against tagging + labeling. I just see labels as more rigid than tags and therefore more reliable as an indicator of what version the APB actually is; or said another way, you can tag an image as `v1.0.0`, `latest`, or `3.7` but what matters w.r.t. the broker knowing what version of the APB tool built you is the label.

djzager (commented):
I like this plan in general to support APB versioning. I am a fan of the APB tool storing the version as a label in the Dockerfile that it builds and the broker using that label to determine the image's version.

rthallisey (approved):

PR #121, opened 2017-09-26
Bug 1497819 - Remove image

djzager:
Remove the image field from generated apb.yml files.

Bug 1497819

This depends on https://github.com/openshift/ansible-service-broker/pull/433 being merged.


src/apb/engine.py
```diff
@@ -623,20 +623,11 @@ def cmdrun_init(**kwargs):
     else:
```
djzager, line 681: At line 619, we split on `/` with:

```
     apb_tag_arr = kwargs['tag'].split('/')
     apb_name = apb_tag_arr[-1]
     if apb_name.lower().endswith("-apb"):
         app_name = apb_name[:-4]
```

As it stands, with these changes, a `apb init <my_org>/my_apb` would ignore the `<my_org>/`. In order to build the image with the organization  like so `apb build --tag <my_org>/my_apb` or `apb build --registry docker.io/<my_org>` for the desired result. Just wanted to point this out to get feedback.

dymurray: Does it make sense here to add org as an option to `apb build`? I would expect to be able to include my org when building using the tooling but --registry docker.io/\<org> isn't that intuitive.

rthallisey:
Before we merge, just want to verify: https://github.com/openshift/ansible-service-broker/pull/433#issuecomment-332934673

djzager:
Once 433 is merged, will need to create a bug for this change (since it only works if the broker ignores the image field).

dymurray (approved):
Thanks for adding the org argument. Looks good to me. We plan to merge this after this sprint I'm assuming?

djzager:
To be honest @dymurray I don't know the answer to that.

jwmatthews:
Once we have a Bug # reference in the description for this PR and have tested we can merge.

djzager:
[(3) Remove image from APB Spec.](https://trello.com/c/siyUDbM0/456-3-remove-image-from-apb-spec)

PR #124, opened 2017-09-27
Initial relist support

eriknelson:
* Will automatically use proper relist with an APB push following
a push (unless otherwise specified with --no-relist)

Adds "apb relist" and will use correct relist following an "apb push".

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -366,20 +366,63 @@ def get_asb_route():
     return asb_route
 
 
-def delete_controller_manager_pod():
-    pod_name = None
+def broker_resource_url(host, broker_name):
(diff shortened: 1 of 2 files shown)


eriknelson:
Want to improve some error handling for likely scenarios (like running against a pre v0.0.21 catalog).

cfchase (commented):
I would think we would want to relist after bootstrap as well?

eriknelson:
@cfchase good point. Will update.

src/apb/engine.py
```diff
@@ -781,6 +828,9 @@ def cmdrun_bootstrap(**kwargs):
 
     print("Successfully bootstrapped Ansible Service Broker")
 
+    import pdb; pdb.set_trace()
```
cfchase: did you mean to remove this?

cfchase (approved):
ACK

PR #129, opened 2017-10-03
Bug 1498185 - Move version declaration into APB spec

dymurray:


dymurray:
This depends upon: https://github.com/openshift/ansible-service-broker/pull/468

shawn-hurley:
Should we add a check in `apb build` to look for the spec version and then we can error if it missing and give an option to skip? I think that might help authors who have updated the apb tool but not the spec?

dymurray:
@shawn-hurley Good suggestion. I error out if the version doesn't exist.

shawn-hurley (approved):

PR #163, opened 2017-11-14
Bump APB Version

djzager:
The asb-modules, specifically `asb_encode_binding`, creates a
secret instead of writing to a file. This breaks backwards compatibility
with older brokers. Therefore, we should bump the APB version string.

See [the
proposal](https://github.com/openshift/ansible-service-broker/blob/master/docs/proposals/prop-apb-gen-creds.md)
for more information.


djzager:
This is no longer needed.

PR #165, opened 2017-11-15
Update documentation

cfchase:
3 main sections now.

Design (high level)
[docs/design.md](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/docupdate/docs/design.md)

Developers Guides (writing apbs)
[docs/developers.md](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/docupdate/docs/developers.md)
[docs/getting_started.md](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/docupdate/docs/getting_started.md)

Tooling (installation and usage)
https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/docupdate/docs/apb_cli.md

PR #166, opened 2017-11-16
update copr links

jmontleon:
I split the ansible-service-broker copr repo into latest and nightly variants. I want to remove the non-suffixed ansible-service-broker repo as it is causing unnecessary load on copr by duplicating most builds. To get there I'm just updating the docs to suggest using the latest repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ There are three tags to choose from:
 
 For RHEL or CentOS 7:
 ```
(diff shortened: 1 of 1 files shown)


dymurray (approved):

cfchase:
Added as a commit on https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/165

PR #176, opened 2017-12-08
Some basic error handling updates and more customization

dymurray:


djzager (approved):
LGTM

docs/apb_cli.md
```diff
@@ -160,6 +160,12 @@ If you are using a namespace other than the default `openshift` namespace to hos
 apb push -o --namespace <namespace>
 ```
 
+This is assuming that the user has the cluster-admin role assigned to the user that is logged in. i.e.:
+```
```
cfchase, line 164: do we want to include an OR `oc login -u system:admin`

src/apb/cli.py
```diff
@@ -212,6 +212,20 @@ def subcmd_push_parser(subcmd):
         help=u'Namespace to push to internal OpenShift registry',
         default=u'openshift'
     )
+    subcmd.add_argument(
+        '--registry-service-name',
+        action='store',
+        dest='reg_svc_name',
+        help=u'Namespace where internal registry service lives',
```
cfchase: help copy paste error from namespace.  Should probably be `Service name where...`

docs/apb_cli.md
dymurray: Can't use system:admin when we require a token unfortunately.

cfchase (approved):
ACK

PR #178, opened 2017-12-12
APB Run Command

djzager:
This PR implements an `apb run` command that works even as a developer to run an APB in a specified namespace. There are some limitations:

- Since apb build, prepare, etc. all work on the current apb project...this does that too. So I have to be "in" the 'hello-world-apb' project in order to `apb-run` that APB.
- APBs that try to create anything above the permissions of a `$namespace:admin` will fail...hard.
- APBs that use `asb_encode_binding` will fail because we aren't passing the POD_NAME and POD_NAMESPACE to the pod we create...I can add this. But I think the right thing from an APB developer perspective is to specifically create a secret when not using the ansible broker. However, I'm not sure how we would communicate that to APBs.

### DIFF
--- a/src/apb/cli.py
+++ b/src/apb/cli.py
@@ -381,6 +381,7 @@ def subcmd_test_parser(subcmd):
     )
     return
 
+
 def subcmd_run_parser(subcmd):
     """ provision subcommand """
     subcmd.add_argument(
@@ -415,6 +416,7 @@ def subcmd_run_parser(subcmd):
         help=u'Name of Dockerfile to build with'
     )
     return
+
 
 def subcmd_relist_parser(subcmd):
     """ relist subcommand """
--- a/src/apb/cli.py
+++ b/src/apb/cli.py
@@ -79,22 +79,7 @@ def subcmd_build_parser(subcmd):
         '--tag',
         action='store',
         dest='tag',
-        help=u'Tag of APB to build'
-    )
-
-    subcmd.add_argument(
-        '--registry',
-        action='store',
-        dest='registry',
-        help=u'Registry prefix of APB to prepend to tag'
-    )
-
-    subcmd.add_argument(
-        '--org',
-        '-o',
-        action='store',
-        dest='org',
-        help=u'Organization of APB to publish to'
+        help=u'Tag of APB to build (ie. mysql-apb or docker.io/username/mysql-apb)'
     )
 
     subcmd.add_argument(
@@ -384,7 +369,15 @@ def subcmd_test_parser(subcmd):
         '--tag',
         action='store',
         dest='tag',
-        help=u'Tag of APB to build'
+        help=u'Tag of APB to build (ie. mysql-apb or docker.io/username/mysql-apb)'
+    )
+
+    subcmd.add_argument(
+        '--dockerfile',
+        '-f',
+        action='store',
+        dest='dockerfile',
+        help=u'Name of Dockerfile to build with'
     )
     return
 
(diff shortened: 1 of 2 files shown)


djzager:
I still need to fix `apb-test`, cleanup the service accounts and role-bindings et al...but I wanted to get some eyeballs on this before I went through that trouble.

src/apb/cli.py
```diff
@@ -387,6 +388,56 @@ def subcmd_test_parser(subcmd):
...
+
+    subcmd.add_argument(
+        '--tag',
+        action='store',
+        dest='tag',
+        help=u'Tag of APB to build'
+    )
+
+    subcmd.add_argument(
+        '--registry',
```
dymurray: Do we need this if we are already prompting for `tag`?

src/apb/cli.py
```diff
@@ -387,6 +388,56 @@ def subcmd_test_parser(subcmd):
...
+        action='store',
+        dest='registry',
+        help=u'Registry prefix of APB to prepend to tag'
+    )
+
+    subcmd.add_argument(
```
dymurray: Same question as above.

src/apb/cli.py
dymurray: I noticed we have the same thing for build... Maybe it's okay. Just feels weird. I wouldn't be opposed to removing it from build either.

src/apb/engine.py
```diff
@@ -464,84 +465,146 @@ def relist_service_broker(kwargs):
...
+    command = ['entrypoint.sh', action, "--extra-vars", json.dumps(parameters)]
+
+    po = create_pod(
+        image=image,
+        name=name,
+        namespace=ns,
+        command=command,
+        service_account=sa
+    )
+
```
dymurray: This might fail flake8 if you don't have 2 newlines.

src/apb/engine.py
```diff
@@ -464,84 +465,146 @@ def relist_service_broker(kwargs):
...
+                }
+            }
+        )
         print("Created Pod")
+        return pod.metadata.name
     except Exception as e:
         print("failed - %s" % e)
 
 
+def run_apb(project, image, name, action, parameters={}):
```
dymurray: +1 I like how simple this function is.

src/apb/engine.py
```diff
@@ -707,6 +770,35 @@ def print_list(services):
...
+def build_apb(project, dockerfile=None, registry=None, org=None, tag=None):
+    if dockerfile is None:
+        dockerfile = "Dockerfile"
+    spec = get_spec(project)
+    if 'version' not in spec:
+        print("APB spec does not have a listed version. Please update apb.yml")
+        exit(1)
+
+    name = "{}{}{}".format(
+        registry + "/" if registry is not None else "",
```
dymurray: What happens if I set `--tag` and `--registry`?

djzager: Looking at how `apb build` works. I took `tag` to be the name of the apb (since if it is not defined we take the name from the apb.yml). Not sure how we want to do this.

djzager:
At this point, there are still some issues with `apb test` that I'm working through...but `apb run` should be done.

src/apb/engine.py
djzager: Fixed.

djzager:
At this point `apb test` runs...but fails for some reason on the only apb for which we have a test playbook. This is worth investigation, but outside of the scope of this PR. 

https://github.com/ansibleplaybookbundle/mediawiki-apb/issues/18 Created

dymurray:
`apb run` worked for me first try! I'm really happy with this. Let's give folks a little more time to look at it but ACK from me.

djzager:
[(3) Experiment with 'apb run' command for running APBs outside of Broker](https://trello.com/c/rZm87qDl/610-3-experiment-with-apb-run-command-for-running-apbs-outside-of-broker)

djzager:
[(3) Experiment with 'apb run' command for running APBs outside of Broker](https://trello.com/c/rZm87qDl/610-3-experiment-with-apb-run-command-for-running-apbs-outside-of-broker)

PR #179, opened 2017-12-13
Add install-broker command

cfchase:


djzager (changes requested):
I am not currently a huge fan of adding this to the apb tool for a couple of reasons. The biggest reason I have against this is that the `apb tool` has some protection against all the various ways of interacting with the broker because we don't deploy it...if we deploy it now we have to try and make it work for all of the various cases (ie. are we making sure a service catalog exists, do we have the authority to give the broker the permissions it needs). The second reason is that I believe deploying the broker, and running apbs other than the current project, is outside the scope of the apb tool. We definitely have a need to create something for discovering + running APBs and I believe this work should go there.

I can be convinced otherwise. Convince me :sunglasses: .

cfchase:
Originally, the use case for adding this to the tooling was because I was creating projects/users/service accounts/roles from the tooling before executing the [asb-installer](https://github.com/ansibleplaybookbundle/asb-installer/pull/1).  Now that we're using the installer for everything, perhaps we can skip the tooling and refer to the [docs](https://github.com/cfchase/asb-installer/tree/install-asb-from-template)

@jwmatthews, I defer to you.  Should we just close this PR as per @djzager?

cfchase:
[(5) Simplify/Consolidate installing broker as an APB](https://trello.com/c/tbnG5P04/594-5-simplify-consolidate-installing-broker-as-an-apb)

rthallisey (changes requested):
I agree with @djzager. I think this is out of the apb tool's scope.

I think it would be cool to do the same thing that we did with the apb tool running in a container.  ```alias asb='docker run --rm --net=host -v $HOME/.kube:/opt/apb/.kube:z -u $(whoami) asb-installer'```. Then we install the broker with ```asb provision -e dockerhuborg=rthallisey```.  I think this is easier to maintain and simpler.

PR #189, opened 2018-01-08
Fix Typos

darahayes:
Hi there, I know this is very nitpicking but this is just a tiny PR to fix some typos.

it's -> short hand for `it is`
its -> implies possession


rthallisey (approved):
It's important for docs to be correct! Thanks for the fix @darahayes.

dymurray (approved):

PR #191, opened 2018-01-10
[WIP] Sort the apb tool functions into Classes of requests

rthallisey:
Break apart the apb tool logic into 3 types of requests so similar logic can share variables and code.
  - Broker Requests
    - apb list
    - apb push
    - apb relist
    - apb bootstrap
    - apb remove
    - apb run
  - Docker Requests
    - apb build
    - apb test
  - Apb Directory Requests
    - apb init
    - apb prepare
    - apb serviceinstance


src/apb/apb_directory/apb_init.py
```diff
@@ -0,0 +1,210 @@
...
+            if not skip[action]:
+                self._write_playbook(project_path, apb_dict, action)
+                if not skip['roles']:
+                    self._write_role(project_path, apb_dict, action)
+
+
+    def apb_init(self):
+        current_path = self.args['base_path']
+        bindable = self.args['bindable']
+        async = self.args['async']
```
mhrivnak, line 154: ``async`` becomes a language keyword in python 3, so it may be worth avoiding that proactively.

src/apb/apb_directory/apb_init.py
```diff
@@ -0,0 +1,210 @@
...
+        skip = {
+            'provision': self.args['skip-provision'],
+            'deprovision': self.args['skip-deprovision'],
+            'bind': self.args['skip-bind'] or not self.args['bindable'],
+            'unbind': self.args['skip-unbind'] or not self.args['bindable'],
+            'roles': self.args['skip-roles']
+        }
+
+        apb_tag_arr = self.args['tag'].split('/')
+        apb_name = apb_tag_arr[-1]
```
mhrivnak, line 165: What happens if the tag has multiple slashes? Do the middle segments get ignored?

src/apb/broker_request/apb_bootstrap.py
```diff
@@ -0,0 +1,26 @@
+import requests
+
+from broker_request import Broker_Request
+
+
+class Apb_Bootstrap(Broker_Request):
+    '''
+    Senf a bootstrap request to the ansible-service-broker.
```
mhrivnak, line 8: s/Senf/Send/

src/apb/broker_request/apb_bootstrap.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def apb_bootstrap(self):
+        url = "%s/v2/bootstrap" % self.route
+        response = requests.request("post", url, verify=self.verify,
+                                    headers=self.headers)
+
+        if response.status_code != 200:
+            print("Error: Attempt to bootstrap Broker returned status: %d" % response.status_code)
+            print("Unable to bootstrap Ansible Service Broker.")
+            raise
```
mhrivnak, line 24: You'll have to specify an exception to raise. Perhaps ``IOError``.

src/apb/broker_request/apb_list.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+    def _print_json_list(self, services):
+        print(json.dumps(services, indent=4, sort_keys=True))
+
+
+    def _print_service(self, service):
+        cmap = ruamel.yaml.comments.CommentedMap()
+
+        if 'name' in service:
```
mhrivnak, line 22: I suggest boiling this down to:

```
for key in ['name', 'id', 'description', ...]:
  if key in service:
    cmap[key] = service[key]
```

mhrivnak (commented):
Here are some initial little suggestions from reading through the code.

src/apb/apb_directory/apb_init.py
dymurray: Since this is just setting `apb_name` the intention was to ignore the middle segments. This is the value we set for the APB's displayName and title of resources. With that said... there is absolutely no validation being done on the `--tag` input parameter so this could blow up.

dymurray:
@rthallisey are you good if we close this for now?

rthallisey:
Ya rewrite is going to happen after bundle-lib is done

PR #192, opened 2018-01-10
Don't append ansible-service-broker suffix without a check

dymurray:


src/apb/engine.py
```diff
@@ -726,8 +726,6 @@ def retrieve_test_result(name, namespace):
 def broker_request(broker, service_route, method, **kwargs):
     if broker is None:
         broker = get_asb_route()
-    else:
-        broker = "%s/ansible-service-broker" % broker
```
dymurray, line 730: From apb push we were already setting broker so this was appending a second suffix.

src/apb/engine.py
```diff
@@ -1047,8 +1045,6 @@ def cmdrun_push(**kwargs):
     blob = base64.b64encode(spec)
     data_spec = {'apbSpec': blob}
     broker = kwargs["broker"]
-    if broker is None:
-        broker = get_asb_route()
```
paolobueno: @dymurray is this removal correct here? maybe this is what's making `--broker` required

paolobueno: Ah no, it seems it's being called from `broker_request()`

dymurray: @paolobueno Yeah there seems to be something that the openshift restclient is complaining about when running `get_asb_route()` from `broker_request()`.

rthallisey (approved):

PR #196, opened 2018-01-11
Bug 1537599 Update the pod_exec call

cfchase:
Update the pod_exec call to use kubernetes 4.0 client to fix "handshake error" and "Upgrade request required".

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -4,6 +4,7 @@ import uuid
 import base64
 import shutil
 import string
+import random
 import subprocess
 import json
 import requests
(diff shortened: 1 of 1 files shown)


cfchase:
@fabianvf This requires an updated [openshift-restclient-python](https://github.com/openshift/openshift-restclient-python) and [kubernetes python client 4.0](https://github.com/kubernetes-incubator/client-python).

@jmontleon the apb.spec will also needs to change to use those dependencies when available.

To test I had to `pip install kubernetes` to get the up-to-date client and then install openshift-restclient-python from the current source.

djzager:
The good news @cfchase is that the 0.4.x `openshift-restclient-python` has been updated to include the kubernetes client-python 4.0.

rthallisey (approved):

djzager (approved):
LGTM

PR #197, opened 2018-01-12
Proposal: apb tool refactor

rthallisey:


mhrivnak:
One aspect that adds substantial noise and overhead to the "engine" today is printing output. I suggest that would be best paired with the other CLI parts. The actions themselves, currently called "commands", would be more like a library that just gets stuff done.

``cli.go`` (or ``cli.py``) would have all of the Subcommand features you listed. It would parse+validate input, determine the correct function to call that will perform real work, and call it with the right arguments. Additionally I'm suggesting that it would receive a return value from that function, and then be **fully responsible** for all output to stdout.

This would isolate output serialization decisions, such as yaml or json or pretty-print or other, which fields to include, etc. to the same part of code that handles input. For example:

``cli.go`` would have a subcommand for ``apb list``. It would parse and validate arguments somehow. It would then call the correct function in ``commands/broker_requests/apb_list.go``. The return value would be a collection of results in native golang structures. Then the subcommand would serialize those results in the appropriate way, based on user input (``-o something``), and print them to stdout.

Thoughts on that?

If we went that direction, we might then consider having a ``commands`` package that holds everything cli.go has plus whatever serialization code there is, and a separate ``actions`` or ``lib`` package (not sure of the best name) that contains the remaining "engine" code (of course organized into classes as proposed).

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,82 @@
...
+for the new utility, but most of the growth was unilaterally in ```engine.py```.
+Resulting in a lot of duplicate code and multiple code execution pathways that
+are making bugs difficult to fix.
+
+This leads me to propose to the community that I think the apb tool should
+undergo some major changes. The focus would be to break apart ```engine.py```
+into a series of classes in order to reuse code and unifiy duplicate code paths.
+
+If the community is interested in changing the apb tool's structure then there
+should be an addition consideration. Most of the future work for the apb tool
```
dymurray, line 17: nit: additional

rthallisey:
@mhrivnak that sounds good to me. I'll make sure your comment is reflected in the proposal.

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+#### base.go
+Base class for all commands
+
+
+#### apb_<command>.go
+Class for apb command logic
+
+
+## Python to Golang discussion
+Changing from Python to Go will benifit the apb execution future work. But,
```
dymurray, line 77: nit: benefit

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+
+#### cli.go
+The ```cli.go``` primary focus is command processing. It gathers and validates
+input, then calls the corresponding subcommand with the appropriate arguments.
+The subcommands will be classified into 3 groups:
+  - Broker subcommands
+  - Docker subcommands
+  - Apb directory subcommands
```
dymurray, line 60: Is it worth adding direct OCP subcommands? I'm thinking of instances across the code where we are creating service accounts and checking for OCP resources. Perhaps it's not really a `subcommand` but more of a client package.

rthallisey: What we can do is initialize the OCP data like the token and OCP functions in the base class ```base.go```. So in that case it would be like a client package.  Or we can have a class for the Cluster (openshift/kubernetes) and initialize that in ```base.go```.

dymurray: +1, makes sense to me.

shawn-hurley:
I would like to suggest that use the [cobra](https://github.com/spf13/cobra) library for developing the commands in go.

eriknelson:
We've been having a lot of conversations in the broker project about breaking the current repo into 3 projects. The relevant one would be a `libapb`, which would directly affect this effort. Fortunately, I think it should make this easier. The basic idea is that the bulk of the actual APB logic, data structures, validation, execution etc., will be vendorable via `libapb`. It is the authoritative project for all things APB. Anyone working with APBs can standardize on top of this library and benefit from upstream feature development and maintenance. It also means `libapb` consumers become lean layers on top of it, only containing their particular concerns. `apb` becomes basically just a cli wrapper for driving `libapb`.

eriknelson:
It would be interesting to continue the discussion around this in light of the past month's work. Since my last comment, APB should be replaced with "Service Bundle", and we actually have a vendorable `bundle-lib` that is effectively what I described as `libapb` thanks to @shawn-hurley's work.

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+This leads me to propose to the community that I think the apb tool should
+undergo some major changes. The focus would be to break apart ```engine.py```
+into a series of classes in order to reuse code and unifiy duplicate code paths.
+
+If the community is interested in changing the apb tool's structure then there
+should be an addition consideration. Most of the future work for the apb tool
+involves facilitating execution of an apb. This work, in python, is going to be
+more difficult than if it were written in go because an apb tool written in go
+can vendor the broker. This would reduce the amount of work to add the features,
+isolate the code maintainace to the broker's code paths, and give the community
```
jmrodri, line 21: TYPO:  maintainace -> maintenance

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+an addition way to test broker code.
```
jmrodri, line 22: TYPO: addition -> additional

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+Finally, my proposal to the community is to change the structure of the apb tool
+in a two ways:
```
jmrodri, line 25: TYPO: a two -> two

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+#### cli.go
+The ```cli.go``` primary focus is command processing. It gathers and validates
+input, then calls the corresponding subcommand with the appropriate arguments.
+The subcommands will be classified into 3 groups:
+  - Broker subcommands
+  - Docker subcommands
+  - Apb directory subcommands
+
+```cli.go``` will expect a return value from the subcommand it called. Based on
+that subcommand, the return value will be processed before its printed to
```
jmrodri, line 63: TYPO: its -> it's

docs/proposals/apb-tool-refactor.md
```diff
@@ -0,0 +1,86 @@
...
+The apb tool's utility has grown from a few build commands to a large diverse
+set of commands that interact with the broker, build containers, and help
+developers build apbs.  The apb tool's code has grown significantly to account
+for the new utility, but most of the growth was unilaterally in ```engine.py```.
+Resulting in a lot of duplicate code and multiple code execution pathways that
+are making bugs difficult to fix.
+
+This leads me to propose to the community that I think the apb tool should
+undergo some major changes. The focus would be to break apart ```engine.py```
+into a series of classes in order to reuse code and unifiy duplicate code paths.
```
jmrodri, line 14: TYPO: unifiy -> unify

jmrodri:
@shawn-hurley @eriknelson is cobra really the best CLI parsing for golang? When I used it for my sm-photo-tool project I wasn't a fan. I found it intrusive. But it's been a year since I last played with it.

jmrodri (changes requested):
typo changes

docs/proposals/apb-tool-refactor.md
jmrodri: I think the cross-section between how the apb tool touches the things written in python and the things written in golang should be weighed. If we make heavy use of external python libraries from say ansible, then it might be helpful to keep it in python. If not, then re-writing to make use of bundle-lib makes sense.

eriknelson:
@jmrodri it feels like the community's default. I don't think that's enough to warrant a blank pass, but there's a lot of precedence and value in using what everybody else is using. I don't have enough experience with it to assert an opinion. I'll give it a try this weekend and weigh-in.

shawn-hurley:
@jmrodri I would like to hear how it was intrusive?I agree with @eriknelson tt does seem to be the standard that most people are using.

dymurray:
@rthallisey do we want to merge this in and continue work on the google doc/second PR we have open?

rthallisey:
@dymurray I forgot about this PR. I'll merge the relevant bits with https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/279.  Then we can continue updating the google doc and I'll pull it all together for one final product.

rthallisey:
work has already begun here: https://github.com/automationbroker/sbcli

PR #207, opened 2018-01-19
Bug 1536687 - Add minishift support

eriknelson:
* Changes to the apb tool to support apb build and push
* Adding a template used to permission users for usage with the apb tool

### DIFF
--- a/docs/apb_cli.md
+++ b/docs/apb_cli.md
@@ -143,10 +143,42 @@ subcommand:
 
 ```
 
+#### Access Permissions
+
+The `apb` tool requires you to be logged in as a tokened cluster user (`system:admin`
+is not sufficient because it does not have a token that can be used for the tool's authentication).
+In addition, there are a number of `RoleBinding`s and `ClusterRoleBindings` that must
+exist to permit the full breadth of the `apb` tool's functions.
+
+The easiest option is to ensure the user has the `cluster-admin` `ClusterRoleBinding`.
+**To be clear, this is effectively cluster root and should only be used in a development setting**.
+
+```
+oc adm policy add-cluster-role-to-user cluster-admin <user>
+oc login -u <user>
+```
+
+If you would like a more strictly permissioned environment, we have an [Openshift Template](templates/openshift-permissions.template.yaml)
+available that can be applied with the following command:
+
+`oc process -f templates/openshift-permissions.template.yaml -p USER=<your_desired_user> | oc create -f -`.
+
+By default, the template will permission the `developer` user. If that is your user, it
+is safe to leave off the `-p` flag, which overrides the default value. Obviously, this
+command must be run by a user with sufficient permissions to create the various roles.
+The `developer` account does not have such permissions. `oc login -u system:admin` should
+be sufficient.
+
 ## Typical Workflows
 
 #### Local Registry
 In order to use the internal OpenShift Docker Registry to source APBs, you must have configured the Ansible Service Broker to use the `local_openshift` type registry adapter. Please see the [config](https://github.com/openshift/ansible-service-broker/blob/master/docs/config.md#local-openshift-registry) section for more information.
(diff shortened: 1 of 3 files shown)


templates/openshift-permissions.template.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: v1
```
djzager, line 4: Documentation on what to do with this? I'm guessing process and `oc create` it?

docs/apb_cli.md
```diff
@@ -147,6 +147,11 @@ subcommand:
 
 #### Local Registry
 In order to use the internal OpenShift Docker Registry to source APBs, you must have configured the Ansible Service Broker to use the `local_openshift` type registry adapter. Please see the [config](https://github.com/openshift/ansible-service-broker/blob/master/docs/config.md#local-openshift-registry) section for more information.
+
+NOTE: If you are working with a minishift cluster and the [ansible-service-broker addon](https://github.com/minishift/minishift-addons/tree/master/add-ons/ansible-service-broker),
+you must first configure your shell to use the minishift Docker daemon. This can
+be done with `eval $(minishift docker-env)`.
```
djzager: I would add the link from minishift here too https://docs.openshift.org/latest/minishift/using/docker-daemon.html

templates/openshift-permissions.template.yaml
djzager: https://github.com/eriknelson/cdk-minishift-utils/blob/asb-update/addons/ansible-service-broker/ansible-service-broker.addon Looks like where this is going to get used. Could you add a comment to point to the project or some breadcumb as to who is using it.

djzager (approved):
LGTM, a couple nits.

dymurray (approved):
I agree with David's comment around documenting the purpose of the template. But other than that this looks great! Thanks a lot Erik.

templates/openshift-permissions.template.yaml
```diff
@@ -0,0 +1,126 @@
...
+################################################################################
+# Broker endpoint access
+################################################################################
+- apiVersion: rbac.authorization.k8s.io/v1beta1
+  kind: ClusterRoleBinding
+  metadata:
+    name: apbtool-access-asb
+  subjects:
+  - kind: User
+    name: ${USER}
```
rthallisey, line 18: Does ```${USER}``` get rendered somewhere?

templates/openshift-permissions.template.yaml
eriknelson: Good idea, will update.

templates/openshift-permissions.template.yaml
eriknelson: It's a template parameter: https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/pull/207/files/432a3249b02b2a3e118df8309aef09e3ac07b219#diff-c16c6fc009826f88518526f6e2c07042R115

`oc process -p USER=<desired_user` will override the default value of `developer`. I thought there might be some interest in this template beyond minishift, so I parametrized it to allow for an override in a different environment. Probably need some docs as @djzager points out, will update.

eriknelson:
This is the actual addon PR as well: https://github.com/minishift/minishift-addons/pull/76, safe to ignore that cdk repo of mine. I was using it before I knew about minishift-addons.

dymurray:
+1 just want to reiterate that this PR worked for me on a Minishift env.

djzager:
Adding doc label.

templates/openshift-permissions.template.yaml
djzager: If it wasn't clear, I was hoping you would also add a comment in here to point to the `minishift-addons` repo. So that when I forget why this file exist I'll be pointed to that project and remember.

PR #209, opened 2018-01-22
Bug 1536963 Updates required to work with openshift 0.4.0

jmontleon:
We're going to need to wait for this to be available at the least.
<fabianvf> jmontleon: just tagged v0.4.0.a1 which should push itself to pypi once the build passes

It's also entirely possible I've missed other places that need updating.

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -433,8 +433,8 @@ def broker_resource_url(host, broker_name):
 def relist_service_broker(kwargs):
     try:
         openshift_config.load_kube_config()
-        token = openshift_client.configuration.api_key['authorization']
-        cluster_host = openshift_client.configuration.host
+        token = openshift_client.Configuration().get_api_key_with_prefix('authorization')
+        cluster_host = openshift_client.Configuration().host
         broker_name = kwargs['broker_name']
         headers = {}
         if kwargs['basic_auth_username'] is not None and kwargs['basic_auth_password'] is not None:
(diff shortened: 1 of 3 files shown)


fabianvf (approved):
Once 0.4.0.a1 is pushed to pypi I think this should all work fine.

djzager (approved):
You probably need a bugzilla for this.

djzager:
@jmontleon I think this is the bug you should use https://bugzilla.redhat.com/show_bug.cgi?id=1536963

karmab (approved):

src/apb/engine.py
```diff
@@ -433,8 +433,8 @@ def broker_resource_url(host, broker_name):
 def relist_service_broker(kwargs):
     try:
         openshift_config.load_kube_config()
-        token = openshift_client.configuration.api_key['authorization']
-        cluster_host = openshift_client.configuration.host
+        token = openshift_client.Configuration().get_api_key_with_prefix('authorization')
```
djzager, line 436: This also exists on line 1092.

djzager (commented):
Looks like one more spot.

cfchase (approved):

src/apb/engine.py
jmontleon: Thanks. Updated the PR message and fixed 1092.

dymurray (approved):

dymurray (commented):
Looks like installing openshift 0.4.0 is failing on Travis. The latest version in pypi is 0.3.4.

djzager:
"<fabianvf> jmontleon, karimb, dzager: openshift 0.4.0.a1 is now up on pypi, pip install 'openshift==0.4.0.a1' should grab it"

I see there is https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions for installing pre-release versions. Not sure if there is a way to use that so we don't lock it down to the alpha version forever.

PR #211, opened 2018-01-23
Add support for minishift to apb container

eriknelson:
This should be pretty much done, but I did have one question about the docs (see the TODO). Does that atomic command actually work without any of the environment config? It's not going to work for minishift, I'm wondering if this script makes that comment obsolete?

Also wondering if the best advice for this is to copy it into your path. There's plenty of ways to get it into your system as a command; I would rather not encourage folks to curl random scripts off the internet.

### DIFF
--- a/docs/apb_cli.md
+++ b/docs/apb_cli.md
@@ -37,29 +37,29 @@ care of the details so they should be easy to deploy.
 
 [Docker](https://www.docker.com/) must be correctly installed and running on the system.
 
+NOTE: If you are working with a minishift cluster and the [ansible-service-broker addon](https://github.com/minishift/minishift-addons/tree/master/add-ons/ansible-service-broker),
+you must first configure your shell to use the minishift Docker daemon. This can
+be done with `eval $(minishift docker-env)`. For more info, see [minishift's documentation for working with its docker registry](https://docs.openshift.org/latest/minishift/using/docker-daemon.html).
+
 #### Running from a container
 
 Pull the container:
 ```bash
 docker pull docker.io/ansibleplaybookbundle/apb-tools
 ```
-
-Create an alias in your `.bashrc` or somewhere else sane for your shell:
-```bash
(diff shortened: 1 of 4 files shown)


jmontleon (approved):
I don't have a better answer. It looks reasonable to me.

I'm not sure on where the script might be packaged or if it's worth. It seems most users of the container are people having trouble installing from source and using Macs or Linux distros besides Fedora or EL.

My only real thought is that maybe we can provide alias and script instructions if it's not confusing. But laying out both may be a little much to digest to people unfamiliar with it.

docs/apb_cli.md
```diff
@@ -37,30 +37,37 @@ care of the details so they should be easy to deploy.
...
+Copy the [apb-docker-run.sh](../scripts/apb-docker-run.sh) script into your `PATH` and
+make sure it's executable:
+
+```
+cp $APB_CHECKOUT/scripts/apb-docker-run.sh $YOUR_PATH_DIR/apb && chmod +x $YOUR_PATH_DIR/apb
 ```
 
 You should be able to start working by running `apb init my_apb`. The first run may take awhile if you did not pull the image.
 
+> TODO: Will the following actually work?
```
djzager: I was originally going to suggest removing this, but I think it makes sense for this to stay.

djzager (approved):
LGTM. I like the idea of putting this logic into the apb-tools container instead of in a `chmod +x` shell script, but I don't have anything better than this at the moment.

docs/apb_cli.md
rthallisey: atomic run is essentially the same thing as docker run, so it should work.

rthallisey: We can remove the ```atomic run``` command. Folks that use atomic will be aware how to run this on atomic since it's s/atomic/docker/.

rthallisey (approved):

PR #220, opened 2018-02-01
Bug 1523252 - Prepend broker route with https if no schema supplied

dymurray:
If the user specifies the `--broker` flag and does not include a schema we should automatically include a schema so we don't see ugly errors.


jmontleon (approved):
ACK

rthallisey (approved):

PR #225, opened 2018-02-07
Bug 1536687 - Prevent script from running as root

eriknelson:
* The apb container script will blow up if the user runs as root, and
its entirely unecessary, so we should prefer least privilege.


djzager (approved):
ACK

dymurray (approved):

PR #226, opened 2018-02-07
Bug 1537599 - Push when running apb test.

cfchase:


djzager (commented):
LGTM. Finishing up something else and I'll test and approve. What's a good example apb to use?

cfchase:
@djzager this is what I was using.  feel free to play with it.  https://github.com/cfchase/test-apb

src/apb/engine.py
```diff
@@ -1232,6 +1232,9 @@ def cmdrun_test(**kwargs):
         kwargs['tag']
     )
 
+    registry = get_registry(kwargs)
```
djzager: If you have a look at the `cmdrun_run` you'll see we set the tag based on the registry like:

```
    apb_project = kwargs['base_path']
    registry = get_registry(kwargs)
    spec = get_spec(apb_project)
    tag = registry + "/" + kwargs['namespace'] + "/" + spec['name']

    image = build_apb(
        apb_project,
        kwargs['dockerfile'],
        tag
    )
    push_apb(registry, tag)
```

I think you will need to do something like that in order for this to work correctly.

djzager (changes requested):
Just need to set the tag based on the registry before you build it in order for it to work correctly. After that I believe this will be ready.

src/apb/cli.py
```diff
@@ -402,6 +402,38 @@ def subcmd_test_parser(subcmd):
         dest='dockerfile',
```
djzager, line 395: While you are at it, if you could remove the `tag` argument please.

djzager (approved):
ACK

dymurray (approved):

eriknelson (changes requested):
I don't think this takes minishift  into account, but I'm not positive since you might be using the same functions a regular push does. Can you double check?

dymurray:
@eriknelson I think it does actually. The minishift logic is handled in this function: https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/a2a180a85f7e387362214a65097efa4550724a77/src/apb/engine.py#L1338

`build_apb` and `push_apb` both call this function when building and pushing. @djzager am I right in that?

eriknelson (approved):

eriknelson:
Confirmed it does use the shared logic. Thanks @dymurray, the other one I was concerned about was the tag, but that's also hidden behind `get_registry`. It's important that the internal minishift registry ip gets used when tagging images, if you're pushing to minishift.

PR #229, opened 2018-02-23
Fix Canary. pip install -U pip setuptools is failing

jmontleon:


djzager (approved):
ACK

eriknelson:
I haven't had any problems building canary without this patch -- I'm...not sure what to make of that, but something very strange is going on with these images and their dependencies.

eriknelson (approved):

PR #230, opened 2018-02-23
Bug 1548543 - Fix minishift SSL Error with latest/downstream

jmontleon:


jmontleon:
From bug:

Description of problem:
apb push
version: 1.0
name: test-apb
description: This is a sample application generated by apb init
bindable: False
async: optional
metadata:
  displayName: test
plans:
  - name: default
    description: This default plan deploys test-apb
    free: True
    metadata: {}
    parameters: []
Finished writing dockerfile.
Building APB using tag: [172.30.1.1:5000/openshift/test-apb]
Error accessing the docker API. Is the daemon running?
Exception occurred! Error while fetching server API version: hostname '192.168.42.253' doesn't match 'localhost'

Version-Release number of selected component (if applicable):
apb-1.1.9

How reproducible:
Always

Steps to Reproduce:
1. Install Minishift
2. Try to do an apb push with the latest, nightly, or downstream images

Actual results:
Exception occurred! Error while fetching server API version: hostname '192.168.42.253' doesn't match 'localhost'

Expected results:
apb push works

Additional info:
The certificate is created with the hostname localhost, which will never work when connecting remotely.

This works properly with the canary image because we get a newer version of backports.ssl_match_hostname when using pip to install dependencies. I did some hacky stuff to downgrade this package on the canary image and saw the same behavior.

This comes from a core RHEL package so I don't think it would  be wise for us to try and update it. We can use assert_hostname=False when setting up the tls connection via python-docker to work around the issue with the older versions.

There appear to be at least one or two changes between 3.4.0.2 and 3.5.0 that change the IP address handling behavior that are likely the reason that this works in newer versions available from canary or on Fedora. As an example:
https://bitbucket.org/brandon/backports.ssl_match_hostname/commits/a8ef5d616d92405a4a74fbcb4bf026cf4d18f030?at=default

eriknelson (approved):

PR #231, opened 2018-02-23
Add network setup script

eriknelson:
* Adds script for creating necessary iptables rules for using containerized apb
* Add note explaining setup-network.sh to docs.


scripts/setup-network.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+iptables -I INPUT -m tcp -p tcp -m multiport --dports 443,8443 -d 192.168.42.0/24 -j ACCEPT
```
djzager: Is minishift guaranteed to take this IP?

djzager (approved):
ACK. One question.

scripts/setup-network.sh
eriknelson: I've never seen it on another subnet, but I can parameterize it to allow someone to override it easily.

PR #245, opened 2018-03-19
Bug 1555390 - Backport fixes/improvements to getting the broker route

dymurray:
* Adds check for looking in the downstream broker namespace
* Prepends/Appends proper suffix/prefix so that `--broker` flag is more usable


jmontleon (approved):
ACK

PR #249, opened 2018-03-24
chore: Minor doc fixes + fix of an error message

rhuss:
I'm just making my way through the excellent getting_started guide, and I hope you don't mind if I collect some minor corrections and maybe questions I had in this pull request.


rhuss:
>  In development mode, it will persist, but usually this namespace would be deleted after successful completion.

Actually, I wonder now how to switch on the 'development mode' because I can't find that temporary namespace after provisioning. So I assume that I'm in 'production mode' (btw, using the getting_started guide with an Minishift setup)

rhuss:
Ah, found:

> In order to use this feature, the Ansible Service Broker you are running must be configured to run in development mode. In the config file, set devbroker to true

However, the link was broken. Not sure whether I picked up the right one now.

Also it looks like the it's named `dev_broker`, not `devbroker`:

```yaml
   broker:
      dev_broker: true
      bootstrap_on_startup: true
      refresh_interval: "600s"
      launch_apb_on_bind: false
      output_request: true
      recovery: true
```

Interestingly, this flag is already set for me, but the temporary namespace didn't survive still.

rhuss:
When adding a DC to the provisioning, then the instruction reads:

> Provision using the Web Console UI.

That's a bit brief. Does this mean I have to unprovision the original service first and the perform the same steps ? Or is there an way to upgrade the provisioned service directly.

docs/getting_started.md
```diff
@@ -707,7 +705,7 @@ apb push
...
+To test an APB you have built without pushing your image to a registry, you can use `apb push`. This command takes the Ansible Service Broker's route as an argument and will push the base64 encoded spec into the list of available APBs for the Broker to deploy. In order to use this feature, the Ansible Service Broker you are running must be configured to run in development mode. In the [config file](https://github.com/openshift/ansible-service-broker/blob/master/etc/example-config.yaml#L44), set `devbroker` to `true`. This enables an endpoint to the broker at `/v2/apb/` that a user can POST APBs to.
```
djzager, line 708: I think this is one of the things that must be updated now that `apb push` uses the internal OpenShift registry by default.

> To test an APB you have built, you can use `apb push`. By default, this command will push the container image to the Local OpenShift Registry, tell the Broker to bootstrap (rebuild the catalog), and bump the relistRequests on the ClusterServiceBroker object (so the APB will be visible in the UI). In order to use this feature, the Broker must be configured for the [Local OpenShift Registry](https://github.com/openshift/ansible-service-broker/blob/master/docs/config.md#local-openshift-registry).

src/apb/engine.py
```diff
@@ -957,7 +957,7 @@ def push_apb(registry, tag):
         openshift_config.load_kube_config()
         api_key = openshift_client.Configuration().get_api_key_with_prefix('authorization')
         if api_key is None:
-            raise Exception("No api key found in kubeconfig. NOTE: system:admin" +
+            raise Exception("No api key found in kubeconfig. NOTE: system:admin " +
```
djzager, line 960: Nice

djzager (approved):
LGTM. A few comments.

rthallisey (approved):

rthallisey:
> Ah, found:
>>    In order to use this feature, the Ansible Service Broker you are running must be configured to run in development mode. In the config file, set devbroker to true
>
>However, the link was broken. Not sure whether I picked up the right one now.
>
> Also it looks like the it's named dev_broker, not devbroker:
```
   broker:
      dev_broker: true
      bootstrap_on_startup: true
      refresh_interval: "600s"
      launch_apb_on_bind: false
      output_request: true
      recovery: true
```
>
> Interestingly, this flag is already set for me, but the temporary namespace didn't survive still.
>

Thanks for the contribution @rhuss! 

To make the temporary namespace survive you can set these options in the broker config:
```yaml
openshift:
  // Always keep the namespace
  keep_namespace: false
  // Only keep on errors
  keep_namespace_on_error: true
```

rthallisey:
> When adding a DC to the provisioning, then the instruction reads:
>
>>    Provision using the Web Console UI.
>>
> That's a bit brief. Does this mean I have to unprovision the original service first and the perform the same steps ? Or is there an way to upgrade the provisioned service directly.
>
To provision the same service again, you'll need to deprovision the existing one and re provision the service with the new resources.  You can make services updatable with the ```updatable: true``` flag in [apb.yml](https://github.com/ansibleplaybookbundle/postgresql-apb/blob/master/apb.yml#L30).

PR #257, opened 2018-04-02
Add example using oc new-app instead of apb push

dymurray:


docs/developers.md
```diff
@@ -418,6 +418,88 @@ etherpad_db_host: "{{ lookup('env','ETHERPAD_DB_HOST') | default('mariadb', true
...
+
+* Step 3: Bootstrap the Automation Broker
+
+This is normally also handled by `apb push` or `apb bootstrap`. I recommend `apb bootstrap` for this step since it will also relist the Service Catalog without you having to wait 5-10 minutes. If you do not have access to do `apb bootstrap`, you can also do the following:
+```
+$ oc get route -n ansible-service-broker
+NAME       HOST/PORT                                           PATH      SERVICES   PORT        TERMINATION   WILDCARD
+asb-1338   asb-1338-ansible-service-broker.172.17.0.1.nip.io             asb        port-1338   reencrypt     None
+
+$ curl -H "Authorization: Bearer $(oc whoami -t)" -k -X POST https://asb-1338-ansible-service-broker.172.17.0.1.nip.io/ansible-service-broker/v2/bootstrap
```
jwmatthews, line 486: Should we call out the privileges required of the logged in user so their token is sufficient to talk to Broker?

jwmatthews:
@akostadinov please take a look at the documentation workaround @dymurray proposes.  Do you see any other issues or concerns we should address for a local developer who does not have docker access?

Our thought is to get a first pass of this working, then come back to fine tune the experience.  
One of things for us to look into further is s2i integration, we've done some work already with a demo of functionality, be good for us to revisit and improve:  https://www.youtube.com/watch?v=ZhLvlpqHI8U

docs/developers.md
dymurray: +1 good call, added a link to our docs.

docs/developers.md
```diff
@@ -418,6 +418,89 @@ etherpad_db_host: "{{ lookup('env','ETHERPAD_DB_HOST') | default('mariadb', true
...
+aXRsZTogTWVkaWF3aWtpIEFkbWluIFVzZXIgKENhbm5vdCBiZSB0aGUgc2FtZSB2YWx1ZSBhcyBB\
+ZG1pbiBVc2VyIFBhc3N3b3JkKQogICAgICAgIHJlcXVpcmVkOiBUcnVlCiAgICAgIC0gbmFtZTog\
+bWVkaWF3aWtpX2FkbWluX3Bhc3MKICAgICAgICB0eXBlOiBzdHJpbmcKICAgICAgICB0aXRsZTog\
+TWVkaWF3aWtpIEFkbWluIFVzZXIgUGFzc3dvcmQKICAgICAgICByZXF1aXJlZDogVHJ1ZQogICAg\
+ICAgIGRpc3BsYXlfdHlwZTogcGFzc3dvcmQK"
+```
+
+* Step 2: Populate the internal OCP registry with our built APB image
+
+This is what is normally handled by `apb push`. In order to build our image without using Docker, we will take advantage of the source-to-image functionality of OpenShift. By default, the Automation Broker is configured to look at the `openshift` namespace for published APBs. The `openshift` namespace is detailed in our documentation as a namespace which exposes it's images/imagestreams to be available to any authenticated user on the cluster. We will take advantage of this by using `oc new-app` in namespace `openshift` to build our image.
```
mhrivnak: s/it's/its/

docs/developers.md
```diff
@@ -418,6 +418,89 @@ etherpad_db_host: "{{ lookup('env','ETHERPAD_DB_HOST') | default('mariadb', true
...
 ```
 
+### Alternative to using `apb push`
+When developing APBs, there are a couple of factors which could prevent the developer from using the full development lifecycle that the APB tooling offers. Primarily these factors are:
+* Developing against an OpenShift/Kubernetes cluster that exists on a remote host
+* Developing APBs on a machine that doesn't have access to the Docker daemon
+
+If a developer meets any of these criteria, then we suggest the following workflow to publish images to the internal OCP registry so that the Automation Broker can bootstrap the image. This section will show you how to do these steps with the APB tooling and without.
+
+* Step 1: Ensure the base64 encoded spec is a label on the Dockerfile
```
mhrivnak: minor suggestion, but I think of this as being "in" the Dockerfile, not "on". It's not a property of the file, but part of its contents.

mhrivnak (changes requested):
Good stuff!

Generally I'll just suggest using "service bundle" instead of "apb" except when you mean the `apb` tool.

akostadinov:
@dymurray , hi, do I understand correctly that one needs to:

- run as regular user at local workstation (at no time root is needed)
- run the `oc` commands with cluster-admin privileges

Just want to make sure I understand correctly. Other than that, I think this looks good. Would be ideal if at some point `apb` tool can work in the one more or the other depending on user choice.

@jianzhangbjz, FYI, maybe we can use more the OpenShift workflow in testing and low number of `apb` functionality tests.

jianzhangbjz:
@akostadinov Yes, you're right. Thank you for your information.

dymurray:
@akostadinov You do *not* need to run as `cluster-admin` in this regard to get `oc new-app` working. You will only need elevated permissions to get the broker commands to work (curl to bootstrap the broker for example). 

Can you clarify what you meant by:
>Just want to make sure I understand correctly. Other than that, I think this looks good. Would be ideal if at some point `apb` tool can work in the one more or the other depending on user choice.

Are you referring to different permissioning levels?

akostadinov:
@dymurray , do you know which permissions would be enough? Perhaps we need do document them.

wrt `work in the one mode or the other depending on user choice` (I had a typo and initially wrote `more` instead of `mode`). I mean that `apb` could use the `new-app` approach instead of user calling `oc new-app ...` in case user request that.

dymurray:
@akostadinov Yes I linked the permission documentation in this PR actually. And I see about your second point. Yes I think its worth filing an issue that we can trivially support this functionality with a new command in the tooling,

djzager (approved):
ACK

dymurray:
@mhrivnak is this good to merge now?

mhrivnak (approved):

PR #258, opened 2018-04-05
Bug 1573714 - PoC token arg support

akostadinov:
Hi, we want to supply authentication token to `apb` as a parameter. It works
with any auth configuration unlike basic auth. Also sometimes kubeconfig
might be in a non-standard location.

I haven't tested the code, just wanted to hear whether this kind of change
would be accepted before spending more time on it.

P.S. maybe we also need a parameter `--kube-config`

### DIFF
--- a/src/apb/cli.py
+++ b/src/apb/cli.py
@@ -570,6 +570,14 @@ def main():
         default=os.getcwd()
     )
 
+    parser.add_argument(
+        '--token',
+        action='store',
+        dest='auth_token',
+        help=u'Specify OpenShift auth token to be used',
+        default=None
+    )
+
     subparsers = parser.add_subparsers(title='subcommand', dest='subcommand')
     subparsers.required = True
 
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -452,6 +452,8 @@ def relist_service_broker(kwargs):
                        base64.b64encode("{0}:{1}".format(kwargs['basic_auth_username'],
                                                          kwargs['basic_auth_password']))
                        }
+        elif kwargs['auth_token'] is not None:
+            headers = {'Authorization': kwargs['auth_token']}
         else:
             headers = {'Authorization': token}
 
@@ -777,6 +779,8 @@ def broker_request(broker, service_route, method, **kwargs):
                        base64.b64encode("{0}:{1}".format(kwargs['basic_auth_username'],
                                                          kwargs['basic_auth_password']))
                        }
+        elif kwargs['auth_token'] is not None:
+            headers = {'Authorization': kwargs['auth_token']}
         else:
             token = openshift_client.Configuration().get_api_key_with_prefix('authorization')
             headers = {'Authorization': token}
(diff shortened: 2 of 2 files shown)


djzager:
I'm interested in this.

jianzhangbjz:
Thank you @akostadinov :+1: @djzager We need this feature to make our automation work for the `apb` tool. I'm glad to test it if you need.

src/apb/cli.py
```diff
@@ -74,6 +74,13 @@ def subcmd_list_parser(subcmd):
         dest='basic_auth_password',
         help=u'Specify the basic auth password to be used'
     )
+    subcmd.add_argument(
```
jianzhangbjz: I think adding the `auth_token` here is not very suitable. The other subcommands, such as `apb bootstrap`, `apb remove`, and so on, also need the `auth_token` when running.

dymurray: +1, I think that `auth_token` could be added to the main list of args. In any command where interaction with the cluster is not used then the token will just be ignored.

src/apb/engine.py
```diff
@@ -452,6 +452,8 @@ def relist_service_broker(kwargs):
                        base64.b64encode("{0}:{1}".format(kwargs['basic_auth_username'],
                                                          kwargs['basic_auth_password']))
                        }
+        elif kwargs['auth_token'] is not None:
```
dymurray, line 455: I notice we have duplicate code here. Thoughts on breaking out this logic to be something like `cluster_request` and then in functions like `relist_service_broker` and `broker_request` we can call this function to instantiate the headers.

Just a nit :)

akostadinov: Agreed there is duplication that's worth dealing with. But I barely know python syntax as well I don't know the `apb` codebase so refactoring anything would require me to put put quiet some time on it and I don't have it atm :confused:

src/apb/cli.py
akostadinov: @jianzhangbjz , very good catch, my bad for not noticing where I put this. Now I moved to the main params.

I still can't get `apb` working locally. Would you mind giving token auth a try?

jianzhangbjz: I'm sorry I missed this message, I will have a try later today.

src/apb/engine.py
rthallisey: I think when we rewrite the apb tooling we can incorporate this. I'm fine with it as is

rthallisey (approved):

djzager (approved):

dymurray (approved):

djzager:
@jianzhangbjz If this is something you would like to see in the 3.10 release, then a Bugzilla will need to be created. If you decide to create a bug, feel free to assign it to me and add it to the conversation here and I'll make sure it gets built. Otherwise, this PR will have to wait until we branch for 3.10 and this work will go into the 3.11 release.

jianzhangbjz:
@djzager I'm so sorry for replying late! Thank you for your support, bug is here: https://bugzilla.redhat.com/show_bug.cgi?id=1573714

PR #259, opened 2018-04-06
Add a command to refresh all the catalog data

rthallisey:
This should help mitigate the issue where the catalog is holding old data.

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -1255,31 +1255,31 @@ def cmdrun_refresh(**kwargs):
     class_output = subprocess.check_output(delete_classes, stderr=subprocess.STDOUT, shell=True)
 
     if "No resources found." in class_output:
-        print "No resources found."
-        print "Run `apb relist` to populate the catalog with clusterserviceclasses"
+        print("No resources found.")
+        print("Run `apb relist` to populate the catalog with clusterserviceclasses")
     else:
-        print "Deleting clusterserviceclasses"
-        if class_output != None and len(class_output) > 0:
+        print("Deleting clusterserviceclasses")
+        if class_output is not None and len(class_output) > 0:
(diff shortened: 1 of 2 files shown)


src/apb/engine.py
```diff
@@ -1250,6 +1250,40 @@ def cmdrun_bootstrap(**kwargs):
...
+        print "Run `apb relist` to populate the catalog with clusterserviceplans"
+    else:
+        print "Deleting clusterserviceplans"
+        if plan_output != None and len(plan_output) > 0:
+            for sp in plan_output.rstrip('\n').split('\n'):
+                delete_plan = "oc delete clusterserviceplan {}".format(sp)
+                output = subprocess.check_output(delete_plan, stderr=subprocess.STDOUT, shell=True)
+                print output
+
+    print "Catalog data has be removed. Relisting data.."
```
djzager: been removed.

src/apb/engine.py
```diff
@@ -1250,6 +1250,40 @@ def cmdrun_bootstrap(**kwargs):
         relist_service_broker(kwargs)
 
 
+def cmdrun_refresh(**kwargs):
+    delete_classes = "oc get clusterserviceclass --no-headers | awk '{ print $1 }'"
```
djzager: Can you use `kubectl` instead? I know that we must shell out since these aren't supported in the openshift/k8s client but  since we don't know how long this version of the apb tool will last it would be nice to at least attempt to be agnostic.

djzager: Same for the others below.

djzager (changes requested):
This should be helpful. Looks good, just a few comments.

dymurray (approved):
LGTM

djzager (approved):
ACK

PR #263, opened 2018-04-17
Update Makefile template

mikenairn:
Don't force the use of the containerised apb tools in Makefile commands, instead apb should be installed locally using one of the recommended approaches https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/apb_cli.md#installing-the-apb-tool


philbrookes:
Awesome :+1:

dymurray (approved):
Makes sense to me. If someone is using the container then it should be aliased so that `apb` just works.

djzager (approved):

PR #267, opened 2018-04-25
Fix RPM Spec

djzager:


djzager:
I removed the sed instruction from the spec and built using `rpmbuild -ba ~/rpmbuild/SPECS/apb.spec` with success.

jmontleon:
This might work... but I think we don't have pip at all in RHEL, so it may still break in a downstream build env.

jmontleon (approved):

jmontleon:
ACK

PR #274, opened 2018-04-30
Bump openshift, no max on websocket

djzager:


jmontleon:
ACK from me. 

Maybe we can try pulling this out too:
https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/Dockerfile-canary#L14

djzager:
Went ahead and bumped the RPM and pulled the check from `Dockerfile-canary`.

PR #279, opened 2018-05-01
APB tool v2 proposal

rthallisey:


jwmatthews:
I think this is a good starting point for the implementation details/structure.  

I'd like to see us flesh this out more with a preliminary draft of capabilities we want to see supported.

I believe the big push to rewriting the CLI in go is to enable the ability to consume APBs outside of the broker, i.e. allow a user to install the cli and then directly deploy applications to their clusters from APBs.

This will allow APBs to be consumed by Brokers and directly from CLI.

How about we collaborate together on a google docs version, once we have some initial consensus there we convert it to markdown and update this PR with that content?

Link to a starting point in google docs:
https://docs.google.com/document/d/1YPkIVGlp-T55FiXHt2h4ccTac7aKJI2VxLCFn8lpyCQ/edit#heading=h.nqovbrswtnmz

jianzhangbjz:
That's cool! I'm glad to see the new version of the apb tool rewriting in Golang. :+1:

rthallisey:
Work has begun here: https://github.com/automationbroker/sbcli

PR #280, opened 2018-05-02
Bug 1572452 - Deprecate --secure flag for SSL verification

dymurray:
If this is undesired by the team please let me know. I figured that this flag is only causing more confusion than anything else and I don't know a single person who uses it. Most developers will be working on a cluster with self-signed certs so I don't expect this to work on majority of developer setups.


djzager (approved):

djzager:
Normally for this stuff we merge after one ACK. In this case, with your question and all, we should get another set of eyes.

PR #285, opened 2018-05-08
Bug 1572452 - Change --secure option to specify CA cert

dymurray:
Looking [here](http://docs.python-requests.org/en/master/user/advanced/) it appears we were using the `verify` flag as a pure boolean.`verify` can also take in a path to a CA cert to verify connections with. I added a new argument `--ca-path` which now allows the user to pass in the path to a custom CA cert. `--secure` will still set a boolean `True` and will expect that the certs are installed via `certifi` or can be grabbed from a public set of certs (check docs for list).

### DIFF
--- a/src/apb/engine.py
+++ b/src/apb/engine.py
@@ -746,6 +746,12 @@ def broker_request(broker, service_route, method, **kwargs):
     if not broker.startswith('http'):
         broker = 'https://' + broker
 
+    if kwargs["verify"] is True and kwargs["cert"] is not None:
+        verify = kwargs["cert"]
+    else:
+        verify = kwargs["verify"]
+    print(verify)
+
     url = broker + service_route
     print("Contacting the ansible-service-broker at: %s" % url)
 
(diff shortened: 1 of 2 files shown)


dymurray:
By `installed ca-bundle` I mean installed via certifi. That is what is supported in 2.16+ of requests.

dymurray:
On a downstream install of openshift-ansible I was able to do:
```
apb list --ca-path /etc/origin/master/ca.crt
```

src/apb/engine.py
```diff
@@ -746,6 +746,11 @@ def broker_request(broker, service_route, method, **kwargs):
     if not broker.startswith('http'):
         broker = 'https://' + broker
 
+    if kwargs["cert"] is not None:
```
djzager, line 749: What's happening here? Not immediately clear to me.

djzager: I read this as users can pass the `--verify` flag to use installed certs, using the `cert` flag to specify a cert forces us to use the specified ca for verification. Is that right?

djzager (approved):
LGTM

src/apb/engine.py
dymurray: That is exactly right. So the following are acceptable:
```
apb list --secure --ca-path <path>
apb list --ca-path <path>
apb list --secure
```

PR #291, opened 2018-05-11
relax python-docker version req

jmontleon:


jmontleon:
Fedora 28 has python2-docker 3.2.1 so we need to relax the version dependency.

I tried apb init, prepare build, push, bootstrap, and relist. They seem to be working

djzager (approved):

dymurray (approved):

PR #292, opened 2018-05-17
APB integration guide for pre existing playbooks

rthallisey:


docs/integration.md
```diff
@@ -0,0 +1,122 @@
...
+There are three things required:
+  1. The production playbooks distinguish between broker APIs
+  2. APB pulls in production playbooks
+  3. APB uses a playbook to call the production playbooks
+
+## Distinguishing Broker APIs
+Since there are multiple APIs an APB expects to plug in to, your production
+playbooks will need to distinguish which action to perform.  The APB will
+provide to your playbooks the variable ```apb_action``` which will be one of
+the APB actions: 'provision', 'deprovision', 'bind', or 'unbind'.
```
mhrivnak: `update` and `test` are also valid options. Hopefully we'll soon have arbitrary action support too, so it may be best to avoid enumerating them here. Maybe just reference the service bundle contract if you want to reference an authoritative list.

docs/integration.md
```diff
@@ -0,0 +1,122 @@
+# Integrating Production Playbooks with APBs
```
mhrivnak: What exactly do you mean by a "production playbook"?

docs/integration.md
```diff
@@ -0,0 +1,122 @@
...
+Using this API, the developer can separate the ansible-playbooks from the APB
+code by pulling in roles from Ansible Galaxy, rpms, or git repos and referencing
+them from the playbook.
+
+There are three things required:
+  1. The production playbooks distinguish between broker APIs
+  2. APB pulls in production playbooks
+  3. APB uses a playbook to call the production playbooks
+
+## Distinguishing Broker APIs
```
mhrivnak: I don't think I'd call this a "broker API". There's nothing broker-specific about it. In fact, it's the APB's entrypoint script that expects the playbooks to be at particular paths. What about just calling it "Actions"?

"Distinguishing Actions

Since there are multiple actions an APB expects to implement, ..."

docs/integration.md
```diff
@@ -0,0 +1,122 @@
...
+container.
+
+```bash
+...
+
+RUN yum install -y production-mysql-playbooks
+```
+
+#### GitHub Source
+Finally, you can install the playbooks right from source by closing from git or
```
mhrivnak: s/closing/cloning/

docs/integration.md
```diff
@@ -0,0 +1,122 @@
...
+
+```yaml
+---
+- name: Provision mysql
+  shell: oc create -f mysql-rc.yaml
+  when: "{{ apb_action }}" == "provision"
+```
+
+## Production Playbook Installation
+Now that the playbooks are separated from the APB repo, the APB needs to install
```
mhrivnak: This is the first mention of a "repo", which could confuse the reader. "What repo? Do I need a specific repo for this?"

This sentence also implies that playbooks were not separate from the APB repo previously. The reader might think they missed a step where these things were supposed to be separated from each other.

Could you replace both sentences with something like this?

"The APB container image needs to install the playbooks, possibly using one of the techniques below."

mhrivnak (changes requested):

docs/integration.md
rthallisey: Playbooks that are already used in production to deploy a service.  A synonym would be pre-existing playbooks.

mhrivnak: I think it might be more natural to call them "Pre-existing Playbooks". Scoping this to "production" makes me think you're differentiating it from dev, test, staging, etc.

docs/pre-existing-playbook-integration.md
```diff
@@ -0,0 +1,130 @@
...
+1. [Pre-existing Playbook Installation](#pre-existing-playbook-installation)
+    1. [Ansible Galaxy](#ansible-galaxy)
+    1. [RPMs](#rpms)
+    1. [GitHub Source](#github-source)
+1. [Running the Pre-existing Playbooks](#running-the-pre-existing-playbooks)
+    1. [Shim Playbook](#shim-playbook)
+
+## Introduction to Playbook Separation
+APBs expect a specific API in the form of playbook names like
+```playbooks/provision.yml``` locally inside the container.  This can prove
```
djzager: You should use a single ` for preformatted text in-line. This is a nit, but definitely impacting the formatting when reviewing the PR.

docs/pre-existing-playbook-integration.md
```diff
@@ -0,0 +1,130 @@
...
+    1. [Ansible 'when' Statement](#ansible-when-statement)
+1. [Pre-existing Playbook Installation](#pre-existing-playbook-installation)
+    1. [Ansible Galaxy](#ansible-galaxy)
+    1. [RPMs](#rpms)
+    1. [GitHub Source](#github-source)
+1. [Running the Pre-existing Playbooks](#running-the-pre-existing-playbooks)
+    1. [Shim Playbook](#shim-playbook)
+
+## Introduction to Playbook Separation
+APBs expect a specific API in the form of playbook names like
```
djzager: I think API here is confusing. We expect playbooks named after OSB actions `provision` being one.

docs/pre-existing-playbook-integration.md
```diff
@@ -0,0 +1,130 @@
...
+```playbooks/provision.yml``` locally inside the container.  This can prove
+problematic when trying to using a pre-existing playbook or well tested
```
djzager: trying to use

djzager (changes requested):
A few nitpicks. I don't have any experience with building APBs based off of existing Ansible projects. My impression is that the reuse should be at the role level, including specific roles into provision, deprovision, etc. but this may be a way to go about it. 

Another option I don't see mentioned is to simply copy the `kubevirt.yml` playbook into the APB container to `actions/provision.yml`. Not sure if that improves anything but I guess there are a good handful of options available to APB writers.

djzager:
I view @fabianvf as a resource when talking about structuring and writing good Ansible. I think that he should have a look at this too.

docs/pre-existing-playbook-integration.md
```diff
@@ -0,0 +1,130 @@
...
+playbooks will need to distinguish which action to perform.  The APB will
+provide to the playbooks with the variable ```apb_action``` for this purpose.
+
+#### Ansible 'when' Statement
+Ansible has the ```when:``` to only run tasks if a condition is met. Add these
+to individual tasks, roles, or playbooks to distinguish actions.
+
+```yaml
+---
+- name: Provision mysql
```
fabianvf, line 41: I would add information to this section about conditional role imports as well as `tasks_from`

rthallisey:
>A few nitpicks. I don't have any experience with building APBs based off of existing Ansible projects. My impression is that the reuse should be at the role level, including specific roles into provision, deprovision, etc. but this may be a way to go about it.
>
For best practices I agree, but it depends how someone has written their playbook so this covers all the bases.

>Another option I don't see mentioned is to simply copy the kubevirt.yml playbook into the APB container to actions/provision.yml. Not sure if that improves anything but I guess there are a good handful of options available to APB writers.
>
That's a possible option.  It depends how the playbook is written, but in this case it would work.

docs/pre-existing-playbook-integration.md
rthallisey: I thought three `'s were required. Good tip.

docs/pre-existing-playbook-integration.md
rthallisey: added

rthallisey:
@djzager @mhrivnak any more comments here? @jwmatthews is this doc in the right place to get added to origin docs?

dymurray (approved):

rthallisey:
further comments?

djzager (approved):

PR #294, opened 2018-05-23
Bug 1581629 - Fix relist to use --ca-path argument

dymurray:
I forgot to add this to the relist command so bootstrap succeeds but the actual relist fails since `verify` was simply being set to true and not to the CA path.


dymurray:
Example of the previous change: https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/src/apb/engine.py#L749

jmrodri (approved):
ACK

djzager (approved):

PR #297, opened 2018-05-29
Add fix for new automation broker namespace

dymurray:


djzager (approved):

jmontleon (approved):

jmontleon:
Thank you @dymurray !

src/apb/engine.py
```diff
@@ -407,7 +403,11 @@ def get_asb_route():
         if 'asb' in route.metadata.name and 'etcd' not in route.metadata.name:
             asb_route = route.spec.host
 
-    url = asb_route + "/ansible-service-broker"
+    if asb_route is None:
+        print("Error finding a route to the OpenShift Automation Broker.")
+        return None
+
+    url = asb_route + "/openshift-automation-service-broker"
```
jianzhangbjz, line 410: @dymurray Why hardcode this `/openshift-automation-service-broker` for request URL? I remember it should be `ansible-service-broker` for the ASB?

jsm84: Indeed... this breaks compatibility with the ASB that ships in OpenShift 3.9. The only way around this is to either downgrade or fork (or wait for OCP 3.10) and there appears to be no easy way to downgrade to apb 1.2 or earlier (as in no older packages available via the yum repositories).

jwmatthews: @dymurray @djzager any thoughts?

PR #308, opened 2018-06-19
Corrected name of ASB access ClusterRole

sabre1041:
Corrected the name of the ClusterRole that is installed into an OpenShift environment for the developer template

### DIFF
--- a/templates/openshift-permissions.template.yaml
+++ b/templates/openshift-permissions.template.yaml
@@ -18,7 +18,7 @@ objects:
     name: ${USER}
   roleRef:
     kind: ClusterRole
-    name: access-asb-role
+    name: asb-access
     apiGroup: rbac.authorization.k8s.io
 ################################################################################
 # apb must be able to view the ansible-service-broker project


djzager:
May I ask how the broker is being deployed where this change is needed?

1. Looking at what is used from the [minishift-addon](https://github.com/openshift/ansible-service-broker/blob/ansible-service-broker-1.2.16-1/templates/deploy-ansible-service-broker.template.yaml#L102), the current name is correct

We are trying to unify (and simplify) all of the ways the broker can be deployed, just curious which way is this targetting?

sabre1041:
@djzager OpenShift Container Platform version 3.7 and 3.9.

https://github.com/openshift/openshift-ansible/blob/master/roles/ansible_service_broker/tasks/install.yml#L63-L69

To provide alignment with the supported product, the minishift addon should probably be updated as well

eriknelson:
@sabre1041 +1, that's been on my TODO stack, unfortunately I'm pretty tied up with catalog work for the next couple weeks.

